PR #3, opened 2026-07-26
Add Chinese and English translations

xiangfeidexiaohuo:


didriksg:
@xiangfeidexiaohuo Thank you, great contribution and much-needed. Merging now.

Heads-up: a big parallel change is landing that rewrites some of the same views. I'll reconcile the localization myself (no rebase needed on your end), and your existing translations are untouched.

It also adds some new strings, which I'll fill in with first-pass Chinese. Those are only my drafts though, so follow-up PRs improving any of the wording, new or existing, are very welcome.

didriksg:
Update: we've since added a batch of new UI (support links, virtual-display editing + custom sizes, and a few section labels) and translated those ourselves so nothing fell back to English.

We kept your term choices (e.g. 内置). A few of the new strings are our best guess rather than yours — for instance "Captures" we rendered as 采集项目 — so if any wording reads awkwardly, PRs to refine the language are very welcome.

Thanks again for kicking this off!

PR #4, opened 2026-07-26
Add on-the-fly display disconnect/reconnect

caicaiks:
# Add on-the-fly display disconnect/reconnect (virtual + physical)

## Summary

Adds BetterDisplay-style **on-the-fly disconnect/reconnect** for displays, in two
independent layers:

1. **Virtual displays** — a per-config connect/disconnect toggle (previously you could only
   create/delete a config), plus a fix that gives each virtual display a stable, unique
   identity so macOS can restore its arrangement across a reconnect.
2. **Physical displays** — true hardware disconnect/reconnect of real monitors on Apple
   Silicon via the private SkyLight API `SLSConfigureDisplayEnabled`, matching BetterDisplay's
   "Disconnect Display" feature. A disconnected display is removed from the layout (identical
   to clamshell) and can be reconnected from the menu.

Also includes one unrelated one-line fix to `build.sh` (see below).

## Motivation

Crisp already had the core virtual-display machinery (`CGVirtualDisplay` create/destroy) but
exposed it only as "create config / delete config" — there was no way to toggle a saved
virtual display on and off without deleting it. And there was no support at all for
disconnecting a **physical** display; the old `DisplayManager.toggleDisplay` was a dead stub
that always returned `false`.

### 1. Virtual display connect/disconnect toggle

- `VirtualDisplayService.setConnected(_:configID:)` — a thin primitive over the existing
  `create()`/`destroy()` that connects/disconnects a **saved** config without mutating the
  configs list. connect = `create()` (system sees an attach); disconnect = `destroy()` (ARC
  releases the `CGVirtualDisplay`, system sees a detach).
- `VirtualDisplayView` — the static "Active" badge on each config row is replaced with a
  switch, with a per-row busy state (a `ProgressView`) while the blocking create/destroy call
  is in flight. Delete is unchanged.

### 2. Stable, unique virtual-display identity (bug fix)

Previously **every** virtual display was created with the same hardcoded
`productID = 0x0001 / serialNum = 0x0001`. macOS keys a display's remembered arrangement and
resolution off `vendorID/productID/serialNum`, so identical values mean two virtual displays
collide and their layout can't be restored reliably on reconnect.

- `VirtualDisplayService.stableIdentity(for:)` now derives a deterministic, effectively-unique
  `(productID, serialNum)` pair from `config.id` (the config UUID), forced non-zero.
- The descriptor `name` now uses the config's name instead of a hardcoded "Crisp Virtual".

### 3. Physical display disconnect/reconnect (Apple Silicon)

- **`Crisp-Bridging-Header.h`** — declares two private SkyLight functions:
  - `SLSConfigureDisplayEnabled(config, display, enabled)` — enable/disable a display inside a
    CG configuration transaction. On Apple Silicon this is a true hardware disconnect.
  - `SLSGetDisplayList(...)` — enumerates **all** displays including disabled ones. Required
    because a disabled display disappears from `CGGetOnlineDisplayList`, so we need this to
    recover its ID for reconnect.
- **`PhysicalDisplayToggleService`** (new) — a `@MainActor` `ObservableObject` singleton:
  - `disconnect(_:)` / `reconnect(uuid:)`, wrapped in
    `CGBeginDisplayConfiguration` → `SLSConfigureDisplayEnabled` →
    `CGCompleteDisplayConfiguration(.permanently)` (the pattern used by Lunar's BlackOut and
    screen_tune).
  - `@Published disconnected: [DisconnectedDisplay]` — a snapshot (UUID/name/resolution) of
    each display we turned off, since a disconnected display is no longer in
    `DisplayManager.displays`. Persisted to `UserDefaults`.
  - Safety: **Apple Silicon only** (gated on `isSupported`), and it **refuses** to disconnect
    a display if doing so would leave zero active displays.
  - `reconcile()` — drops records for displays that came back online.
  - `reapplyOnWake()` — re-disconnects displays that macOS re-enabled after sleep/wake.
  - Reconnect re-resolves the current `CGDirectDisplayID` by UUID, since IDs can be reassigned.
- **`PhysicalDisplayToggleView`** (new) — two integrated UI pieces:
  - `DisconnectDisplayRow` — a "Disconnect Display" row shown inside each display's detail
    view (below "Set as Main Display"), gated on Apple Silicon and hidden for the last screen.
  - `ReconnectDisplaysSection` — a "Disconnected" section under the main display list that
    lists disconnected displays with a Reconnect action (they no longer have their own row).
- **`DisplayManager`**:
  - `refreshDisplays()` calls `PhysicalDisplayToggleService.shared.reconcile()`.
  - The dead `toggleDisplay` stub is replaced by `disconnectDisplay(_:)` that delegates to the
    service.
- **`AppDelegate.onWake`** — after the existing 2s WindowServer-settle delay, calls
  `reapplyOnWake()` so a sleep/wake cycle doesn't silently re-enable a disconnected display.
- **`project.yml`** — `OTHER_LDFLAGS` marks the two `SLS*` symbols as resolved-at-runtime
  (`-Wl,-U,...`). SkyLight is already loaded in-process via AppKit, so this avoids hard-linking
  a private framework path.

### Design decision: no auto-disconnect on launch

On relaunch the persisted list is loaded **for display only** — the app never blacks out a
screen by itself at startup. The user reconnects (or ignores) from the "Disconnected" section.
Only the sleep/wake path re-applies a disconnect (via `reapplyOnWake()`).

### Unrelated fix

- **`build.sh`** — the archive-output `grep -E` filter contained `** ARCHIVE`, an invalid
  regex (`**`) that made BSD grep abort with `repetition-operator operand invalid` and swallow
  the real xcodebuild error. Changed to `ARCHIVE`. Not part of the display feature, but small
  and it unblocks reading real build errors.

## Platform / requirements

- **Physical** disconnect/reconnect requires **Apple Silicon + macOS 13+**. On Intel the API
  does not perform a true disconnect, so the feature is hidden there.
- **Virtual** display changes work everywhere the existing feature does.
- Uses private, undocumented APIs (`CGVirtualDisplay`, `SLSConfigureDisplayEnabled`,
  `SLSGetDisplayList`). Not App Store compatible — consistent with Crisp already being
  distributed unsigned. Field names/behavior may change across major macOS versions.

## Files changed

New:
- `Crisp/Services/PhysicalDisplayToggleService.swift` (238 lines)
- `Crisp/Views/PhysicalDisplayToggleView.swift` (117 lines)

Modified:
- `Crisp/Crisp-Bridging-Header.h` — SkyLight declarations
- `Crisp/Services/VirtualDisplayService.swift` — `setConnected`, `stableIdentity`
- `Crisp/Views/VirtualDisplayView.swift` — connect/disconnect toggle
- `Crisp/Services/DisplayManager.swift` — `reconcile()` call, `disconnectDisplay`
- `Crisp/Views/DisplayDetailView.swift` — `DisconnectDisplayRow`
- `Crisp/Views/MenuBarView.swift` — `ReconnectDisplaysSection`
- `Crisp/App/AppDelegate.swift` — `reapplyOnWake()` on wake
- `project.yml` — `OTHER_LDFLAGS` for SkyLight symbols
- `.gitignore` — ignore the `Crisp-bin` dev-loop artifact
- `build.sh` — unrelated grep fix

## Testing

- Builds clean with `xcodebuild ... CODE_SIGNING_ALLOWED=NO` (full Xcode) — **BUILD
  SUCCEEDED**, no warnings in the new files. Also builds via the Command Line Tools `swiftc`
  path in `docs/BUILDING.md`.
- Manually verified on Apple Silicon (M-series, macOS 15.7) with two external displays:
  - Disconnect removes a display from the layout (true hardware disconnect, not just dimming);
    it appears in the "Disconnected" section and Reconnect restores it.
  - Layout is restored on reconnect.
  - Disconnect is refused / hidden for the last remaining active display.
  - Virtual display toggle connects/disconnects a saved config without deleting it.

### Reviewer note / to re-verify

- Sleep/wake `reapplyOnWake()` and cross-relaunch persistence were implemented against the
  verified disconnect primitive but should get one more pass on the reviewer's hardware.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ DerivedData/
 Crisp.xcodeproj/
 Crisp.dmg
 .superpowers/
+Crisp-bin
--- a/Crisp/App/AppDelegate.swift
+++ b/Crisp/App/AppDelegate.swift
@@ -222,6 +222,9 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
                 // touching display state.
                 try? await Task.sleep(nanoseconds: 2_000_000_000)
                 dm.refreshDisplays()
+                // Re-disconnect any physical displays macOS re-enabled on wake.
+                await PhysicalDisplayToggleService.shared.reapplyOnWake()
+                dm.refreshDisplays()
                 try? await Task.sleep(nanoseconds: 500_000_000)
                 for display in dm.displays {
                     // Apply software brightness factor first so GammaService
--- a/Crisp/Crisp-Bridging-Header.h
+++ b/Crisp/Crisp-Bridging-Header.h
@@ -64,6 +64,21 @@ typedef int CGSConnectionID_t;
 extern CGError CGSConfigureDisplayMode(CGSConnectionID_t connection, CGDirectDisplayID display, uint32_t modeID);
 extern CGSConnectionID_t CGSMainConnectionID(void);
 
+// MARK: - SkyLight Private API (physical display disconnect/reconnect, Apple Silicon)
+
+// Enable/disable an individual display inside a CGBeginDisplayConfiguration transaction.
+// On Apple Silicon (macOS 13+) disabling performs a true hardware disconnect (identical to
+// clamshell): the display is removed from CGGetOnlineDisplayList / CGGetActiveDisplayList.
+extern CGError SLSConfigureDisplayEnabled(CGDisplayConfigRef config,
+                                          CGDirectDisplayID display,
+                                          bool enabled);
+
+// Enumerates ALL displays including ones disabled via SLSConfigureDisplayEnabled. Required to
+// recover a disabled display's ID for reconnect, since CGGetOnlineDisplayList omits it.
+extern CGError SLSGetDisplayList(uint32_t maxDisplays,
+                                 CGDirectDisplayID *displays,
+                                 uint32_t *displayCount);
+
 // MARK: - IOAVService Private API (Apple Silicon DDC)
 
 typedef void * IOAVServiceRef;
--- a/Crisp/Resources/Localizable.xcstrings
+++ b/Crisp/Resources/Localizable.xcstrings
@@ -1,6 +1,76 @@
 {
   "sourceLanguage" : "en",
   "strings" : {
+    "Disconnect Display" : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "断开显示器"
+          }
+        }
+      }
+    },
+    "Disconnected" : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "已断开"
+          }
+        }
+      }
+    },
+    "Display configuration failed (CGError %@)." : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "显示器配置失败（CGError %@）。"
+          }
+        }
+      }
+    },
+    "Display not found." : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "未找到显示器。"
+          }
+        }
+      }
+    },
+    "Physical display disconnect requires Apple Silicon (macOS 13+)." : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "断开物理显示器需要 Apple 芯片（macOS 13+）。"
+          }
+        }
+      }
+    },
+    "Reconnect" : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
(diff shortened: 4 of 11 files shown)


didriksg:
Thanks for this, @caicaiks! The physical disconnect/reconnect feature is great. I tested it end to end (disconnect, reconnect, last-display guard, sleep/wake, reboot) with no issues.

One thing before merging: the virtual-display parts overlap with work already on `main` (the `0x0001` identity fix is already there), so those changes in `VirtualDisplayService.swift` and `VirtualDisplayView.swift` would conflict rather than add.

I'd love to merge just the physical feature, with credit to you. Either you re-scope the PR down to the disconnect feature and I squash-merge it, or I graft the physical part onto `main` keeping you as author and close this PR pointing at the merge. Either way you're credited. Which do you prefer?

Thanks again!

caicaiks:
hi @didriksg Thanks for the heads-up — I've resolved the conflicts.

didriksg:
Thanks @caicaiks! I reconstructed the change as a single clean commit on top of current `main` (the branch had drifted a fair bit behind and picked up some unrelated virtual-display churn), keeping you as the author. Disconnect and reconnect, plus the sleep/wake reapply, all tested clean on Apple Silicon. Great addition, thank you for the contribution!

PR #11, opened 2026-07-28
Fix stale color profile subtitle after switching profiles

shaw-baobao:
## Summary

- Share the active color profile name between the display detail row and the profile picker through a binding.
- Update the row subtitle immediately after ColorSync applies a new profile successfully.
- Preserve the existing subtitle and selection when profile application fails.

## Root cause

DisplayDetailView loaded the subtitle when its display ID task ran, while ColorProfileView managed the picker selection locally. A successful profile change therefore updated the picker state without refreshing the parent subtitle.

## Testing

- Ran a full-project Swift type check.
- Built and launched Crisp 1.2.0 locally on macOS 27.
- Switched an external display to the Mi Monitor profile and confirmed that the row subtitle updates immediately.

### DIFF
--- a/Crisp/Views/ColorProfileView.swift
+++ b/Crisp/Views/ColorProfileView.swift
@@ -5,6 +5,7 @@ import SwiftUI
 /// checkmark on the active profile.
 struct ColorProfileView: View {
     @ObservedObject var display: DisplayInfo
+    @Binding var activeProfileName: String
     @State private var profiles: [ICCProfile] = []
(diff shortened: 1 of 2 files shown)


didriksg:
Thanks so much for this, @shaw-baobao! 🙏 I tested it locally and the color-profile subtitle now updates correctly right after switching profiles. Clean, well-scoped fix, and much appreciated. Merging it in now, thanks for contributing to Crisp!

PR #13, opened 2026-07-30
Fix/ddc avservice matching

caicaiks:
## What & why
  External-display brightness was broken on Apple Silicon in three ways, plus a build-tooling convenience:

  - **Adjusting one display changed another.** The ARM64 DDC path matched each `IOAVService` (DDC channel) to a display by walking the IORegistry parent chain for `DisplayVendorID`/`DisplayProductID` — keys that don't exist on Apple
  Silicon (identity lives in a sibling `dispextN` node's `DisplayAttributes.ProductAttributes`). Matching always failed and fell back to sorted-`CGDirectDisplayID` index assignment, which pairs channels in the wrong order. Replaced with
  a depth-first proximity walk + vendor/product/serial matching.
  - **Upper half of the slider did nothing (DELL).** `arm64Read`/`intelRead` returned reply bytes 6–9 as the VCP `max` without validating the frame. Monitors that ack the read but return stale EDID/null bytes gave a bogus `max` (e.g.
  8824 instead of 100), so `percent × max` pushed the top of the slider past the panel's real range. Now the DDC/CI reply frame is validated (source `0x6E`, opcode `0x02`, no-error, VCP echo, `max > 0`); on mismatch `max` falls back to
  the standard 100.
  - **Flicker while dragging (HP).** The coalescing writer recursed with no spacing, firing DDC writes back-to-back; a fast drag (~60/sec) floods the I2C bus. Writes are now paced to ~20/sec (50ms MCCS floor) while still applying the
  latest value.
  - Also: a failed DDC *read* no longer forces the gamma/software fallback (availability is driven by *write* success), and added a `Makefile` wrapping `dev.sh` + the release/DMG scripts.

  ## How tested
  - Builds locally via `./dev.sh` (swiftc, Command Line Tools only); binary swapped into the installed app and re-signed.
  - Verified on a dual-display M4 Pro: **DELL U2720QM** (vendor `0x10ac`, product `0x41c0`) + **HP Z27k G3** (vendor `0x220e`, product `0x36d3`).
  - IORegistry probe confirmed the old path found no `DisplayVendorID`/`DisplayProductID` keys and mapped the two channels in reverse; the new proximity + identity match maps each display to its own `IOAVService`.
  - DDC probe confirmed both monitors return no valid VCP read reply (0/30 across settle times) while writes succeed, so the new validation rejects the garbage `max` and falls back to 100.

  ## Checklist
  - [x] N/A — no new user-facing strings
  - [x] N/A — no UI change (backend DDC fixes)

### DIFF
--- a/Crisp/Services/BrightnessService.swift
+++ b/Crisp/Services/BrightnessService.swift
@@ -218,14 +218,13 @@ final class BrightnessService: @unchecked Sendable {
                     self.ddcMaxBrightness[displayID] = result.max
                     self.ddcAvailableLock.unlock()
                     Task { @MainActor in display.brightness = brightness }
-                } else {
-                    // DDC read returned nil; mark unavailable
-                    self.ddcAvailableLock.lock()
-                    if self.ddcAvailable[displayID] == nil {
-                        self.ddcAvailable[displayID] = false
-                    }
-                    self.ddcAvailableLock.unlock()
                 }
+                // A failed/ignored read does NOT mean DDC is unavailable: many monitors
+                // accept brightness *writes* but never answer *reads* (they ack the I2C
+                // transaction with stale/null bytes, now rejected by DDCService). Leaving
+                // availability undetermined lets the write path decide — marking it false
+                // here would wrongly force the gamma/software fallback on a display whose
+                // hardware backlight control works fine, just showing a stale slider value.
             }
         }
     }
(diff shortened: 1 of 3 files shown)


didriksg:
Thanks @caicaiks, really solid fix. 🙏 Tested it here with no issues. This should resolve the multi-monitor DDC targeting where adjusting one display could drive another (#14), plus the reply-frame validation that fixes the compressed brightness range. Merging now for 1.3.0.

PR #24, opened 2026-08-04
Fix HiDPI mode filtering for portrait displays

shaw-baobao:
## What & why

Rotated portrait displays could lose their usable HiDPI modes because the native aspect ratio was inferred from the largest backing buffer, which could belong to a landscape HiDPI mode. The resolution menu also used landscape-only minimum dimensions.

This change:
- derives the panel aspect from its largest unscaled timing before falling back to scaled modes;
- applies orientation-independent minimum dimensions;
- keeps opposite-orientation modes out of the resolution menu and smooth-scaling slider;
- adds focused geometry regression tests and a `make test` entry point.

## How tested

- `make test` — 3 tests passed
- `make compile`
- `./scripts/release.sh v0.0.0-ci` — universal arm64 + x86_64 DMG built
- Physical 90° external display: 12 eligible portrait HiDPI modes enumerated; switched from 1440×2560 to 1080×1920 with a 2160×3840 backing buffer, then restored the original mode

## Checklist

- [x] Added/updated `Crisp/Resources/Localizable.xcstrings` for any new user-facing strings (N/A: no new strings)
- [x] Screenshot or short clip for any UI change (N/A: no visual or layout change)

### DIFF
--- a/Crisp/Models/DisplayMode.swift
+++ b/Crisp/Models/DisplayMode.swift
@@ -87,8 +87,10 @@ struct DisplayMode: Identifiable, Equatable {
         // clean scaled resolutions (1600x900, 2048x1152, full 2560x1440 refresh set, ...) that
         // CGS carries without any override, so we can offer and apply them directly like BetterDisplay.
         let knownIDs = Set(modes.map { $0.id })
-        let nativeAR = rawModes.max(by: { $0.pixelWidth * $0.pixelHeight < $1.pixelWidth * $1.pixelHeight })
-            .map { Double($0.pixelWidth) / Double($0.pixelHeight) } ?? 0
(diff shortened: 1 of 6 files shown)


bang0223 (approved):

bang0223 (approved):

didriksg:
Hardware-verified on a physically rotated Dell U2412M (1200x1920 portrait) alongside an AOC Q27G3XMN (2560x1440 landscape) and the built-in display:

- Before: the Dell's Resolution section showed "No display modes available".
- After: portrait modes are listed (1200x1920 default, 1000x1600 / 840x1344 / 800x1280 in both HiDPI and non-HiDPI groups), and mode switching works.
- No regression on the landscape AOC or the built-in display.

Thanks for the fix!

PR #34, opened 2026-08-09
i18n: fix interpolation-based localization failures and add missing zh-Hans strings

xiangfeidexiaohuo:
Fix three bugs where string interpolation prevented LocalizedStringKey
from matching defined translations:

- BrightnessSliderView: split ternary into three literal branches so
  "Brightness control mode: DDC hardware" and "...Software emulation"
  translations take effect instead of being looked up as "...: %@".
- MenuBarView: use Text(verbatim:) + NSLocalizedString to compose the
  display state accessibility label from individually localized
  fragments (", main display", ", expanded", ", collapsed").
- DisplayModeListView: replace string-interpolated accessibilityLabel
  with explicit NSLocalizedString lookups so both the label and the
  ", selected" suffix are localized.

Add missing zh-Hans keys to Localizable.xcstrings:
- "Show Volume Sliders" -> "显示音量滑块"
- "Speaker volume" -> "扬声器音量"
- "DDC" -> "DDC"
- ", main display" -> "，主显示器"
- ", selected" -> "，已选中"

## Checklist
- [ ] Screenshot or short clip for any UI change

### DIFF
--- a/Crisp/Resources/Localizable.xcstrings
+++ b/Crisp/Resources/Localizable.xcstrings
@@ -322,6 +322,26 @@
         }
       }
     },
+    ", collapsed" : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "，已折叠"
+          }
+        }
+      }
+    },
+    ", expanded" : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "，已展开"
+          }
+        }
+      }
+    },
     ", main display" : {
       "localizations" : {
         "zh-Hans" : {
--- a/Crisp/Resources/Localizable.xcstrings
+++ b/Crisp/Resources/Localizable.xcstrings
@@ -322,6 +322,26 @@
         }
       }
     },
+    ", main display" : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "，主显示器"
+          }
+        }
+      }
+    },
+    ", selected" : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "，已选中"
+          }
+        }
+      }
+    },
     "%@, expanded" : {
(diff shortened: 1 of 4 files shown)


didriksg (changes requested):
Thanks, the diagnosis is right (interpolated `LocalizedStringKey` coalesces to `%@` keys that never match the catalog) and most of the fixes are good. Two issues to fix before merge:

1. **`DisplayModeListView.swift` (selected accessibility label): double comma, regresses English.** The new `", selected"` catalog entry already carries its own leading comma (`", selected"` / `"，已选中"`), and the code inserts another literal `", "` in front of it, so VoiceOver output becomes `"Full Brightness, , selected"` in English and `"Full Brightness, ，已选中"` (ASCII comma followed by full-width comma) in zh-Hans. Please drop the extra literal `", "` so the fragment supplies its own separator.

2. **`MenuBarView.swift` (display row label): `", expanded"` / `", collapsed"` still fall back to English.** The code now calls `NSLocalizedString(", expanded")` and `NSLocalizedString(", collapsed")`, but those bare keys were never added to `Localizable.xcstrings` (only the `%@, expanded` / `%@, collapsed` variants exist, used elsewhere). So for zh-Hans these fragments still render untranslated, the same bug class this PR fixes. Please add `", expanded"` and `", collapsed"` entries (e.g. `"，已展开"` / `"，已折叠"`).

The `", main display"` fragment, the `BrightnessSliderView` DDC fix, and the `Show Volume Sliders` / `Speaker volume` / `DDC` additions all check out. Happy to merge once the two items above are addressed.

didriksg:
Thanks again for the contribution! Since it's been a few days, I went ahead and pushed the two remaining review items directly to your branch (via maintainer edits) so we can get this merged:

- Dropped the extra literal `", "` before the `", selected"` fragment in `CheckmarkRow`, since the fragment carries its own separator.
- Added the missing `", expanded"` / `", collapsed"` entries to `Localizable.xcstrings` (`"，已展开"` / `"，已折叠"`).

That resolves both points from my review, so I'll merge this now. Thanks for the fix and the zh-Hans additions!

didriksg (approved):
Both review items are addressed (pushed directly to the branch), CI is green after syncing with main. Merging.

PR #35, opened 2026-08-10
fix: render display-default refresh rate as 60hz

YuriNachos:
## What & why
- `CGDisplayMode.refreshRate` returns `0` for a display's default timing. `DisplayMode.refreshRateString` rendered that as `\"-- Hz\"`, contradicting the `DisplayMode.refreshRate` contract documented right on the stored property — *\"Refresh rate in Hz (0 means display default, shown as 60)\"* — and reading as broken next to real rates in the resolution slider and refresh-rate list.
- Render `0` (and any non-positive value) as the `60Hz` default, matching how System Settings presents the default timing and honouring the existing doc. Whole-number (`60Hz`) and NTSC fractional (`59.94Hz`, `47.95Hz`) formatting is unchanged.
- Extracted the formatting into a pure `RefreshRateFormat` helper so it can run under a headless `XCTest`: the rest of `DisplayMode` reaches private CGS APIs through the bridging header the test target does not carry (same extraction route the DDC matcher already uses).

## How tested
- `make test` — all 21 tests pass (6 new in `RefreshRateFormatTests`: zero/negative default, whole-number, near-whole collapse, NTSC two-decimal, cinema rounding)
- `make compile` — clean swiftc -swift-version 5 build
- N/A — pure decision logic, headless XCTest (no hardware/monitor verification needed)

## Checklist
- [x] Builds locally (`make test` / `make compile`)
- [x] N/A — pure decision logic, no new user-facing strings (the only label change is `-- Hz` → `60Hz` for the display-default timing)
- [x] N/A — no UI change

### DIFF
--- a/Crisp/Models/DisplayMode.swift
+++ b/Crisp/Models/DisplayMode.swift
@@ -29,13 +29,7 @@ struct DisplayMode: Identifiable, Equatable {
     }
 
     var refreshRateString: String {
-        guard refreshRate > 0 else { return "-- Hz" }
-        let rounded = refreshRate.rounded()
-        // Whole-number timings render clean ("60Hz"). NTSC fractional timings keep two
-        // decimals ("59.94Hz", "47.95Hz") so they don't collapse onto the neighbouring
-        // whole rate and show up as duplicate rows, matching System Settings.
-        if abs(refreshRate - rounded) < 0.01 { return "\(Int(rounded))Hz" }
-        return String(format: "%.2fHz", refreshRate)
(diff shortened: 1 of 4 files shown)


didriksg:
Thanks for this! Kind of funny in hindsight: the doc comment has said "0 means display default, shown as 60" since 1.0, and the code just never did it. Checked the extracted formatter against the old one and it's identical for real rates, so this only changes the 0 case, and the tests pin it all down nicely. Merged.

PR #36, opened 2026-08-10
test: cover display-mode geometry edges (720p floor, square, all-hidpi fallback)

YuriNachos:
## What & why
- `DisplayModeGeometry` ships only three tests, all on the portrait-menu path. The boundaries its other branches depend on were untested:
  - the **inclusive 720p eligibility floor** and how it splits across the short side (`>= 720`) and long side (`>= 1280`) — exactly-720p must pass, a sub-720 short side *or* a sub-1280 long side (e.g. 4:3 1024x768) must not;
  - **square (orientation-neutral) modes** — a 1:1 mode matches both a landscape and a portrait reference so the resolution menu still offers it under rotation;
  - the **all-HiDPI `nativeAspect` fallback** — when no 1x timing is reported (an Apple Retina panel in CG, where every mode has `pixelWidth > width`), the function falls back to the full set and derives the aspect from the largest mode.
- This adds twelve headless tests that pin each behaviour, mirroring the `DDCServiceMatcherTests` style (each test names the mutation it kills). No source change — the logic is correct, this locks it in.

## How tested
- `make test` — all 27 tests pass (12 new in `DisplayModeGeometryEdgeCaseTests`; the existing 15 still pass)
- `make compile` — clean swiftc -swift-version 5 build
- N/A — pure decision logic, headless XCTest (no hardware/monitor verification needed)

## Checklist
- [x] Builds locally (`make test` / `make compile`)
- [x] N/A — pure decision logic, no new user-facing strings
- [x] N/A — no UI change

### DIFF
--- a/CrispTests/DisplayModeGeometryEdgeCaseTests.swift
+++ b/CrispTests/DisplayModeGeometryEdgeCaseTests.swift
@@ -1,0 +1,120 @@
+import XCTest
+
+/// Headless edge-case coverage for `DisplayModeGeometry`.
+///
+/// `DisplayModeGeometry` is compiled directly into this test target (see `project.yml`
+/// sources). The sibling `DisplayModeGeometryTests` pins the portrait-menu behaviour; the
+/// cases here pin the boundaries it doesn't reach — the 720p eligibility floor, square
+/// (orientation-neutral) modes, and the all-HiDPI `nativeAspect` fallback — so each branch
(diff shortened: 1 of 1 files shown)


didriksg:
Nice one, thanks! Went through these against the implementation and they all hold up. Especially glad to see the all-HiDPI fallback covered, that path had nothing on it before. Merged.

PR #38, opened 2026-08-11
fix: persist gamma adjustments by stable display uuid, not displayID

### LINKED ISSUE #32: Color temperature (gamma) applied to wrong display after reboot on multi-monitor setups
### Description

On a dual **external** monitor setup, per-display **Color Temp** (Image Adjustment / software gamma) does not reliably stick across a full reboot, and can be **applied to the wrong display**.

Concrete example:

1. Leave the primary external display’s color temperature unchanged (neutral).
2. Raise color temperature only on the secondary external display.
3. Reboot the Mac (full restart, not just sleep/wake).
4. After login / Crisp starts:
   - The **primary** external display now has the color temperature that was set on the **secondary**.
   - The **secondary** external display’s color temperature is effectively **unchanged** (still neutral / not the value just set).

Related friction while using Color Temp + presets on the same multi-monitor setup:

- Saving **Presets** does not capture color temperature (and other image-adjustment / gamma fields). Updating a preset after manually tuning Color Temp also does not record those values. If this is intentional scope, it would help to document it; if users expect presets to include image adjustment, that is a separate feature gap.

This is still reproducible on **Crisp 1.4.0**.

### Steps to reproduce

1. Connect **two external** displays (multi-monitor).
2. Open Crisp → expand each external display → Image Adjustment.
3. Keep **Display A** (main / primary external) Color Temp at neutral.
4. Set **Display B** Color Temp away from neutral (e.g. warmer / higher). Confirm the tint is only visible on B.
5. Optionally collapse the section / close the panel so state is persisted.
6. **Restart the Mac** fully.
7. After boot, observe both externals’ Color Temp / white balance without reopening the sliders first.

### Expected behavior

- Each display restores the color temperature (and other gamma adjustments) that was last saved **for that physical display**.
- Display A remains neutral; Display B keeps the warmer setting.

### Actual behavior

- Display A receives Display B’s previous color temperature.
- Display B does not show the adjustment that was set before reboot.

Sleep/wake survival improved in 1.4.0 (#25); this report is specifically about **full reboot** and **cross-display swap** on multi-external setups.

### Environment

- **App:** Crisp **1.4.0**
- **macOS:** current stable / recent (dual external monitors)
- Two external displays managed via Crisp Image Adjustment (Color Temp)

### Suspected cause (from source)

`GammaService` persists adjustments under:

```text
crisp.GammaService.savedAdjustment.<CGDirectDisplayID>
```

`DisplayInfo` already documents that macOS may reassign `CGDirectDisplayID` across reconnects / wake, and exposes a more stable `displayUUID` (with vendor/model/serial fallback). Brightness-key selected displays already persist by **UUID**, but gamma / Color Temp still keys by volatile **display ID**.

On reboot, if display IDs are reshuffled:

- Physical display A may get the old ID that had B’s saved gamma → A gets B’s color temp.
- Physical display B may get a new ID with no saved state → B looks unchanged / neutral.

`reapplyIfNeeded(for:)` then faithfully re-applies the wrong dictionary entry to the wrong panel. That matches the observed A←B swap after restart.

#25 fixed post-wake transfer-table clobber (re-apply timing / ColorSync / reconfig). It does not appear to migrate gamma keys to `displayUUID`, so reboot identity remap can still mis-assign settings.

### Suggested directions (non-prescriptive)

- Persist gamma / Color Temp (and any other per-display image adjustment) by **`displayUUID`** (or vendor+model+serial), same pattern as brightness-key selection.
- On load, migrate old `crisp.GammaService.savedAdjustment.<displayID>` keys when a live display’s current ID still matches a legacy key.
- When reapplying after boot / reconfiguration, resolve “which saved adjustment” via stable identity, not the current numeric ID alone.
- Optionally document that **Presets** currently store resolution / brightness / arrangement only—not Color Temp / gamma—so users do not expect presets to round-trip image adjustment.

### Related

- #25 — color temperature lost after sleep/wake (fixed in 1.4.0 for wake re-apply; different symptom, same subsystem; that report already noted displayID vs UUID for persistence keys).
- #14 — multi-monitor brightness cross-talk (identity pairing; fixed earlier).

Happy to test a build that keys gamma by UUID, or provide more detail if useful.

didriksg:
## What

On a dual-external monitor setup, `GammaService` persisted Color Temp / other image adjustments under `crisp.GammaService.savedAdjustment.<CGDirectDisplayID>`. macOS can reassign `CGDirectDisplayID` across a full reboot, so after restart the saved adjustment could land on the wrong physical display: A gets B's tint, B goes back to neutral.

This does the same thing the brightness-key selected-display persistence already does: key by `DisplayInfo.displayUUID` instead of the volatile ID.

## Changes

- `GammaService.saveState` / `loadSavedState` / `clearSavedState` / `reapplyIfNeeded` now take the `DisplayInfo` itself and persist/read under `GammaPersistenceKey.uuidKey(for: display.displayUUID)`, not the raw `CGDirectDisplayID`. `apply`, `reapply`, `hasActiveAdjustment`, `applyIdentity`, `resetSingleDisplay` stay keyed by `CGDirectDisplayID` since those only touch the in-memory, session-scoped `activeAdjustments` dict and the live CoreGraphics calls, which need the real id anyway.
- Added `GammaService.migrateLegacyStateIfNeeded(for:)`, called from `DisplayManager.refreshDisplays()` right after the display list is rebuilt (app launch, reconnect, wake), before anything reapplies a saved adjustment. For each live display, if its *current* id still has a legacy `crisp.GammaService.savedAdjustment.<id>` entry, that entry is copied to the display's UUID key and the legacy key is removed. It never overwrites an adjustment already saved under the UUID key, and it never touches a legacy key whose id has no live display right now, that display might just be disconnected, and guessing would repeat the exact bug this fixes.
- Extracted the key construction and the migration decision (which live displays have a matching legacy entry) into a pure `GammaPersistenceKey` enum (`Crisp/Models/GammaPersistenceKey.swift`), same pattern as `DDCServiceMatcher`/`DisplayModeGeometry`: no `UserDefaults` access inside it, so it's testable headlessly. Wired into the `CrispTests` target via `project.yml` sources (not `@testable import Crisp`, same reasoning as the existing pure-type tests). Added `CrispTests/GammaPersistenceKeyTests.swift` covering key construction and the migration mapping, including the dual-external "only the display with a matching legacy id migrates, to its own uuid" case and the "stale legacy entry with no live display is left alone" case.
- Presets still don't store gamma/Color Temp, that's the separate, already-documented gap the issue also mentions. Didn't touch it here.

## Migration behavior

- Runs once per display-list rebuild (`DisplayManager.refreshDisplays()`), so on every app launch, reconnect, and wake.
- Per live display: if `crisp.GammaService.savedAdjustment.<its current id>` exists, it gets copied to `crisp.GammaService.savedAdjustment.uuid.<its uuid>` (unless that UUID key already has a value) and the legacy key is deleted.
- A legacy key whose id doesn't belong to any currently connected display is left in `UserDefaults` untouched. It's orphaned but harmless, and there's no safe way to know which physical display it used to belong to.

## Verified locally

- `make compile` succeeds clean, no new warnings or errors from any of the changed files (checked the two pre-existing warnings in `GammaService.swift`/`DisplayManager.swift` predate this diff).
- `xcrun swiftc -parse` on the new pure type and its test file, both parse cleanly.

## Could not verify locally

- This machine only has the Command Line Tools, no full Xcode, so `make test` / `xcodebuild` don't run here (confirmed: `swiftc -typecheck` on any `CrispTests/*.swift` file, old or new, fails with "no such module XCTest", not specific to this change).
- No multi-monitor hardware here, so I couldn't reproduce the original bug or verify the actual reboot/reconnect behavior end to end. @cicae, you offered to test a build keyed by UUID, if you're up for it, that'd be the real verification this needs. I'd appreciate a check that Color Temp now survives a reboot with each display keeping its own value.

Fixes #32

### DIFF
--- a/Crisp/App/AppDelegate.swift
+++ b/Crisp/App/AppDelegate.swift
@@ -258,7 +258,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
                         // Apply software brightness factor first so GammaService
                         // can read the up-to-date factor when it re-applies its formula.
                         BrightnessService.shared.reapplySoftwareBrightnessIfNeeded(for: display)
-                        GammaService.shared.reapplyIfNeeded(for: display.displayID)
+                        GammaService.shared.reapplyIfNeeded(for: display)
                         // Re-apply any custom resolution that macOS may have reset on wake
                         ResolutionService.shared.reapplySavedModeIfNeeded(for: display.displayID)
                     }
--- a/Crisp/Models/GammaPersistenceKey.swift
+++ b/Crisp/Models/GammaPersistenceKey.swift
@@ -1,0 +1,53 @@
+import Foundation
+import CoreGraphics
+
+/// Pure key-construction and migration-decision logic for `GammaService`'s persisted
+/// per-display adjustments (issue #32: macOS can reassign `CGDirectDisplayID` across
+/// reboots/reconnects, so keying storage by displayID alone can apply a saved color
+/// temperature to the wrong physical display on a dual-external setup). Mirrors the
+/// identity mechanism already used for brightness-key display selection:
+/// `DisplayInfo.displayUUID`.
+///
+/// Owns no `UserDefaults` access itself, so it can be exercised headlessly from an
+/// `XCTestCase` (same route as `DDCServiceMatcher`).
+enum GammaPersistenceKey {
+    private static let base = "crisp.GammaService.savedAdjustment"
+
+    /// Current storage key: keyed by the stable `DisplayInfo.displayUUID`, which
+    /// survives a `CGDirectDisplayID` reassignment.
+    static func uuidKey(for uuid: String) -> String {
+        "\(base).uuid.\(uuid)"
+    }
+
+    /// Legacy storage key from before issue #32's fix: keyed by the volatile
+    /// `CGDirectDisplayID`. Still consulted once, at migration time, for a display
+    /// whose current id happens to still match what was saved under it.
+    static func legacyKey(for displayID: CGDirectDisplayID) -> String {
+        "\(base).\(displayID)"
+    }
+
+    /// One legacy entry that should move to its display's stable UUID key.
+    struct MigrationTarget: Equatable {
+        let legacyKey: String
+        let uuidKey: String
+    }
+
+    /// Pure decision core for the legacy-to-UUID migration: which currently online
+    /// displays have a legacy, displayID-keyed saved adjustment that should move under
+    /// the stable UUID key.
+    ///
+    /// A display's legacy entry migrates only when its *current* `CGDirectDisplayID` is
+    /// in `legacyDisplayIDsWithSavedState` (the caller determines this by checking
+    /// storage for exactly the ids the live displays currently have, nothing else).
+    /// Legacy keys belonging to a displayID that isn't live right now are never
+    /// considered here: guessing at those is exactly the bug being fixed (a stale
+    /// entry silently reassigned to the wrong physical display after a reboot).
+    static func migrationTargets(
+        liveDisplays: [(id: CGDirectDisplayID, uuid: String)],
+        legacyDisplayIDsWithSavedState: Set<CGDirectDisplayID>
+    ) -> [MigrationTarget] {
+        liveDisplays
+            .filter { legacyDisplayIDsWithSavedState.contains($0.id) }
+            .map { MigrationTarget(legacyKey: legacyKey(for: $0.id), uuidKey: uuidKey(for: $0.uuid)) }
+    }
+}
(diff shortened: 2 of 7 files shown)


didriksg:
Two updates on the "could not verify locally" section since this merged:

- The dev machine has full Xcode now, so the test suite runs locally as well as in CI. All 52 tests pass, including this PR's GammaPersistenceKey suite.
- The end-to-end check happened: with both externals connected, color temperature survived a reboot with each display keeping its own value. So the fix is verified on real multi-monitor hardware, not just in the unit tests.

@cicae no need to test anything on your end anymore, but if things behave for you after the next release, a note here would still be appreciated.

PR #53, opened 2026-08-15
fix: show menu panel when opened from a full-screen app's Space

### LINKED ISSUE #52: Menu panel does not appear when opened from a full-screen app's Space
Clicking the menu bar icon while a full-screen app is active (menu bar revealed on hover) highlights the icon, but the panel never appears.

**Cause**: `MenuPanel.collectionBehavior` in AppDelegate lacks `.fullScreenAuxiliary`, so the panel can't join the full-screen app's Space and gets placed on the desktop Space instead. Every other overlay window in the app (EDR, notch, arrangement) already uses `.canJoinAllSpaces + .fullScreenAuxiliary`.

crayonlu:
Adds `.canJoinAllSpaces` and `.fullScreenAuxiliary` to `MenuPanel.collectionBehavior` so the panel can join a full-screen app's Space. Previously the click only highlighted the icon; the panel was placed invisibly on the desktop Space.

Fixes #52

### DIFF
--- a/Crisp/App/AppDelegate.swift
+++ b/Crisp/App/AppDelegate.swift
@@ -973,7 +973,10 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
         p.hasShadow = false
         p.animationBehavior = .none
         p.isReleasedWhenClosed = false
-        p.collectionBehavior = [.transient, .ignoresCycle]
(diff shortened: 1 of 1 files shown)


didriksg:
Good catch, thanks! The notch/EDR overlays already set these two flags, the menu panel just never got the memo. Will merge once CI goes green.

Issue #54, opened 2026-08-15
点击平滑缩放授权之后没反应


newment:
点击平滑缩放授权之后没反应, 重新启动也不行, 无法打开平滑缩放

didriksg:
Thanks for the report. I'm answering in English since I unfortunately don't read Chinese, but feel free to keep writing in Chinese on your end.

Quick background on what that password prompt does: turning on Smooth scaling writes a display override file (that's the admin part), then blinks the display off and on for about a second so macOS picks it up. On some setups macOS simply ignores that override file, and then it looks exactly like what you're describing: you enter the password and nothing changes, even after a reboot.

I can't tell from the report alone whether that's what's happening here, so a few things would help me a lot:

- Which Mac (model and chip, e.g. MacBook Air M2 or Intel MacBook Pro), and which macOS version?
- Which monitor, and how is it connected? HDMI, DisplayPort, USB-C, or through a dock or adapter? DisplayLink docks in particular are known trouble.
- After you enter the password: does the screen go black for a second? And does the Smooth scaling switch stay on, or flip back off?
- If you can, run this in Terminal and paste the output. It tells me whether the override file was actually written:

  ```
  ls -R /Library/Displays/Contents/Resources/Overrides/
  ```

With that I should be able to tell where it's going wrong.

newment:
font{
        line-height: 1.6;
    }
    ul,ol{
        padding-left: 20px;
        list-style-position: inside;
    }


    
    
[...]

PR #60, opened 2026-08-16
fix: apply smooth scaling immediately on single-display setups

### LINKED ISSUE #58: Smooth scaling toggle silently does nothing on single-display setups (Mac mini)
### Description

On a machine with a single active display (Mac mini or Studio with one monitor, MacBook in clamshell with one external), toggling Smooth scaling looks like it does nothing: the admin prompt appears, the plist is written or removed, but the switch snaps back to its previous state. The change then applies "by surprise" on the next reboot or cable replug, at which point the switch is stuck in the new state the same way.

Diagnosed in #54 (Mac mini M4 Pro, single Xiaomi monitor over HDMI).

### Root cause

`userToggleSmooth` writes/removes the override plist, then calls `softReconnect` to blink the display so macOS re-reads it (`DisplayModeListView.swift`). `softReconnect` refuses when the target is the only active display (`wouldLeaveNoActiveDisplay` guard in `PhysicalDisplayToggleService.swift`), to avoid stranding the user on a black screen if the re-enable fails. The caller discards `softReconnect`'s Bool result, so the refusal is silent: no re-enumeration happens, `smoothModesPresent` still reports the old state, and the optimistic switch snaps back with no explanation.

### Fix plan

Two layers, shipped together:

1. **Honest UX fallback.** Stop discarding `softReconnect`'s return value. When the blink is refused or fails, keep the plist change and show a hint ("Saved. Unplug and replug the monitor cable, or restart your Mac, to apply") instead of silently snapping the switch back. New user-facing string, needs zh-Hans in `Localizable.xcstrings`.

2. **Allow the blink on a single display, with a safety net.** Permit `softReconnect` for this re-enumeration blip even when the target is the only active display (the guard stays for user-initiated disconnects):
   - Before disabling, persist a "soft-reconnect pending" marker (display UUID) in UserDefaults; clear it once the re-enable succeeds. On app launch, if the marker is present and that display is SLS-disabled, re-enable it. This heals a crash mid-blink without touching displays other apps (Lunar, BetterDisplay) may have disabled intentionally.
   - Keep the existing 3-retry re-enable loop; on final failure, attempt to re-enable every disabled display that is not in the user's intentional `disconnected` set, then surface the layer-1 hint.
   - Worst case after all layers is a reboot, which resets WindowServer display state anyway.

The ~1s blank on the only display is already what the UI warns about ("briefly flashes the screen").

### Affected

Every single-display desktop setup, which is most Mac mini and Mac Studio users. Multi-display setups are unaffected (the blink runs there today).

didriksg:
## What & why
Fixes #58 (diagnosed in #54). On single-active-display machines (Mac mini with one monitor, clamshell with one external), the Smooth scaling toggle wrote the override plist but `softReconnect` silently refused to blink the only display, so nothing re-enumerated and the switch snapped back with no explanation. The change then applied by surprise on the next reboot or replug.

Two layers:
1. **Honest fallback**: `userToggleSmooth` no longer discards `softReconnect`'s result. When the blink is refused or fails, a neutral persistent hint explains that the change is saved and needs a cable replug or restart. New string localized in en + zh-Hans.
2. **Blink allowed on a single display, with a safety net**: the `wouldLeaveNoActiveDisplay` refusal is removed from `softReconnect` only (kept for user-initiated disconnects). A dead-man marker (display UUID in UserDefaults) is set before the disable and cleared after a successful re-enable; startup recovery re-enables only that marked display if a crash stranded it. If all 3 re-enable retries fail, every unintentionally disabled display (not in the intentional `disconnected` set) is re-enabled as a last resort, and the layer-1 hint still shows.

Review question: the retry-exhausted fallback sweeps all unintentionally disabled displays, while startup recovery touches only the marked one. Intentional asymmetry per the fix plan, but worth a second look.

## How tested
`make check` passes (swiftlint --strict, xcodebuild test with warnings-as-errors, localization key completeness). NOT yet verified on real single-display hardware: the blink/re-enable path on a sole active display needs a live test (Mac mini, or MacBook in clamshell with one external).

## Checklist
- [x] Builds locally
- [x] Added `Localizable.xcstrings` entries for the new user-facing string (en + zh-Hans)
- [ ] Screenshot of the new hint row pending live testing

### DIFF
--- a/Crisp/Resources/Localizable.xcstrings
+++ b/Crisp/Resources/Localizable.xcstrings
@@ -1628,6 +1628,16 @@
         }
       }
     },
+    "Saved. Unplug and replug the monitor cable, or restart your Mac, to apply." : {
+      "localizations" : {
+        "zh-Hans" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "已保存。请拔掉并重新插入显示器连接线，或重启 Mac，以使更改生效。"
+          }
+        }
+      }
+    },
     "Saving…" : {
       "localizations" : {
         "zh-Hans" : {
--- a/Crisp/Services/DisplayManager.swift
+++ b/Crisp/Services/DisplayManager.swift
@@ -236,6 +236,10 @@ class DisplayManager: ObservableObject {
         // Keep the physical-disconnect list honest: drop any record whose display came back
         // online (re-plugged, or macOS re-enabled it).
         PhysicalDisplayToggleService.shared.reconcile()
+        // Runs on every refresh, but only ever does anything right after launch: a soft
+        // reconnect (smooth-scaling toggle) leaves a marker that only a mid-toggle app
+        // crash/force-quit would still find set, and it clears itself on first check.
+        Task { await PhysicalDisplayToggleService.shared.recoverStrandedSoftReconnect() }
         // A physical unplug bypasses disconnect()'s last-screen guard: internal disabled via
         // Crisp + external cable pulled = zero active displays, all black. Bring one back.
         PhysicalDisplayToggleService.shared.restoreIfNoActiveDisplay()
--- a/Crisp/Services/DisplayManager.swift
+++ b/Crisp/Services/DisplayManager.swift
@@ -236,9 +236,10 @@ class DisplayManager: ObservableObject {
         // Keep the physical-disconnect list honest: drop any record whose display came back
         // online (re-plugged, or macOS re-enabled it).
         PhysicalDisplayToggleService.shared.reconcile()
-        // Runs on every refresh, but only ever does anything right after launch: a soft
-        // reconnect (smooth-scaling toggle) leaves a marker that only a mid-toggle app
-        // crash/force-quit would still find set, and it clears itself on first check.
+        // Runs on every refresh, but is a cheap no-op unless a soft reconnect's dead-man
+        // marker is set (mid-toggle crash, or a re-enable that failed outright). The service
+        // skips it while a live soft reconnect is mid-blink, so this can't race the toggle's
+        // own retry loop even though the blink's reconfig events land here mid-toggle.
         Task { await PhysicalDisplayToggleService.shared.recoverStrandedSoftReconnect() }
         // A physical unplug bypasses disconnect()'s last-screen guard: internal disabled via
         // Crisp + external cable pulled = zero active displays, all black. Bring one back.
--- a/Crisp/Services/DisplayManager.swift
+++ b/Crisp/Services/DisplayManager.swift
@@ -238,8 +238,8 @@ class DisplayManager: ObservableObject {
         PhysicalDisplayToggleService.shared.reconcile()
         // Runs on every refresh, but is a cheap no-op unless a soft reconnect's dead-man
         // marker is set (mid-toggle crash, or a re-enable that failed outright). The service
-        // skips it while a live soft reconnect is mid-blink, so this can't race the toggle's
-        // own retry loop even though the blink's reconfig events land here mid-toggle.
+        // skips any display whose soft reconnect is still mid-blink, so this can't race a
+        // toggle's own retry loop even though the blink's reconfig events land here mid-toggle.
         Task { await PhysicalDisplayToggleService.shared.recoverStrandedSoftReconnect() }
         // A physical unplug bypasses disconnect()'s last-screen guard: internal disabled via
         // Crisp + external cable pulled = zero active displays, all black. Bring one back.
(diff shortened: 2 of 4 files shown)


didriksg:
An adversarial review pass on the first commit found two defects, fixed in b496f64:

1. **Crash recovery defeated itself.** `recoverStrandedSoftReconnect` cleared the dead-man marker unconditionally (via `defer`), even when its single re-enable attempt failed, permanently abandoning a stranded display after one try. It also ran mid-toggle: the blink's own removeFlag reconfig event fires `refreshDisplays`, which invoked recovery concurrently with the live retry loop. Now: recovery mirrors softReconnect's ladder (3 retries, then the sweep), clears the marker only once the display is verifiably online or gone, and is skipped while a soft reconnect is in flight. A useful side effect of marker-gating instead of launch-gating: a display left stranded by an outright failed toggle gets retried on later refreshes too, not just at next launch.

2. **The replug hint could be wrong or lost.** When the retry-exhausted sweep effectively completes the blink (disable landed, sweep's re-enable landed), the override gets re-read, the modes re-enumerate, and `softReconnect`'s `false` return would have shown a 'replug to apply' hint that isn't true; in that same path the controller gets rebuilt, so the hint would land on an orphaned instance anyway. The hint is now gated on the settled switch state not matching the requested one, so it appears exactly when the toggle visibly failed to land. (The reviewer suggested plumbing the hint through `DisplayManager` like `pendingResolutionExpandUUID`; the ground-truth gate handles every reachable case without new plumbing, since the only path that rebuilds the controller is also the path where the hint would be incorrect.)

Full `make check` green on both commits. Live single-display blink still needs on-hardware verification.

didriksg:
Second review round (on b496f64) confirmed both prior fixes are sound for a single blink, but caught that the dead-man marker and in-flight guard were process-wide singletons while two soft-reconnects can genuinely overlap (manual toggle on display A while `autoEnableHiDPIIfNeeded` blinks a freshly plugged display B): the second blink clobbered the first's marker, and a crash could strand a display with no marker left for recovery to find. Fixed in the latest commit by keying both per display UUID: marker is now a list, in-flight is a UUID set, recovery walks every stranded marker (with the same burst guard `restoreIfNoActiveDisplay` has), and the sweep skips displays whose own blink is live. `make check` green.

didriksg:
Live test on a MacBook + AOC external caught two bugs no static review found, fixed in 3b7cbc4:

1. **Clamshell Sleep races the blink.** With the lid closed, the sole active external held the machine awake; the blink's disable removed it and macOS entered Clamshell Sleep mid-toggle (pmset log: `Entering Sleep state due to 'Clamshell Sleep'` seconds after the toggle). The display was left SLS-disabled and undiscoverable until re-enabled out-of-band. The sole-display blink is now desktop-only: on machines with a battery, softReconnect refuses and the replug hint shows instead. Mac mini and Studio, the actual #58 audience, keep the blink; lid-open laptops are unaffected (two active displays).

2. **`SLSConfigureDisplayEnabled(true)` can report success while the display stays disabled** (observed around the sleep transition). Both the toggle's retry loop and marker recovery cleared the dead-man marker on that false positive. Recovery now trusts only re-enumeration: a re-enable counts once the display verifiably reappears in the online list (polled ~1s), otherwise the marker survives for the next refresh or launch.

Remaining live verification: the blink on an actual single-display desktop (Mac mini), which no laptop can simulate. Clamshell + lid-open paths are now hardware-tested.

didriksg:
Final live-testing round, superseding my earlier comment about refusing the blink on portables: `ed779a1` makes the blink work in clamshell too.

**What live testing on a MacBook + AOC Q27G3XMN (HDMI, clamshell) established:**
- A `PreventSystemSleep` assertion does NOT stop Clamshell Sleep: the machine entered DarkWake one second after the sole display was disabled, with the assertion held.
- A live virtual display DOES: with a throwaway CGVirtualDisplay online, the same blink ran with no sleep event and the display verified back in one attempt.
- The re-enumeration resets the display to macOS's default mode (180Hz panel came back at 144Hz), so the exact pre-blink mode is now captured and restored after a verified re-enable.

**Final mechanism:** on portables, when the target is the sole active display, softReconnect holds a minimal virtual display (1080p, filtered from the UI via the shared virtual vendor ID, fixed identity so macOS's what-to-show prompt doesn't recur) for the blink's duration. Desktops need no guard. The replug hint remains only for Intel and for the rare guard-creation failure.

**Verified by hand, both directions, in clamshell:** no sleep, switch holds, refresh rate preserved. Known cosmetic cost: windows briefly migrate to the invisible guard while the real display is off and reshuffle when it returns; inherent to keeping the machine awake through a sole-display blink.

Mac mini remains untested live but is the strictly easier case (no clamshell rule, no guard needed).

didriksg:
Live-tested on my MacBook in clamshell with an AOC Q27G3XMN over USB-C to DisplayPort: toggling smooth scaling in both directions now applies immediately with a single blink, no replug needed.

PR #71, opened 2026-08-21
Fix external brightness delays with multiple displays

### LINKED ISSUE #72: External brightness changes can be delayed by a slow DDC display
### What happened

With two external displays connected, brightness changes in Crisp can occasionally take several seconds or tens of seconds to reach the monitor. When one display has a slow or unresponsive DDC/CI channel, brightness changes for the other healthy display are delayed as well.

The slider should respond immediately, and each external display should process DDC brightness changes independently so one failing channel cannot block another monitor.

### Steps to reproduce

1. Connect two AOC U32N10 external displays.
2. Open Crisp and expand both external display controls.
3. Adjust brightness repeatedly until one display enters the intermittent slow-DDC state.
4. While that display is still processing the DDC request, adjust brightness on the other display.
5. Observe that the second display can wait behind the first request instead of updating independently.

### Crisp version

1.5.0

### macOS version

macOS 26.5.1 (25F80)

### Mac model and chip

MacBook Pro (Mac17,8), Apple M5 Pro

### Monitor(s) and how they are connected

Two AOC U32N10 4K displays connected as separate external display links; one landscape and one rotated 90 degrees. macOS System Information does not report the cable or adapter transport for either display.

### Anything else

A direct channel probe reproduced the difference clearly: one IOAVServiceReadI2C request failed after 12.131 seconds, while the other display completed the same operation in about 10 ms. Crisp used one global serial DDC queue, so the slow request blocked unrelated external-display brightness work.

A fix is proposed in PR #71.

DEOWL-kan:
## Summary

- Match Apple Silicon DDC channels by stable CoreDisplay/IORegistry location before falling back to display identity or traversal order.
- Run DDC operations on a serial queue per display, so one slow I2C channel cannot block every external monitor.
- Coalesce rapid brightness changes to the latest target and apply an immediate software preview while DDC capability is unknown or slow.
- Reject stale DDC reads, writes, and callbacks after newer user input or a display topology change.

## Root cause

Crisp previously routed every DDC transaction through one global serial queue. On the reproduced setup, one external display could block an I2C read for about 12 seconds, which delayed brightness changes for the other display as well. Identical monitors could also be paired by unstable traversal order, while reconnect callbacks could update state belonging to a newer display topology.

## Changes since the review

Rebased onto current `main` (`4275339`). The planning docs and the `.gitignore` line are gone from the branch, and both regressions you found are fixed:

1. **Topology invalidation is scoped to the channels that actually changed.** `DisplayManager` now fingerprints each external ID as vendor/product/serial plus its CoreDisplay location, and only invalidates the IDs whose fingerprint changed or disappeared. Dragging display A's slider survives display B replugging, a new display appearing, and a rearrangement in System Settings (a `.movedFlag` with no identity change now invalidates nothing). The diff itself is `DDCTopologyChange` in `Models/`, covered by `DDCTopologyChangeTests` — including the exact case from your review.

2. **The three-failure latch is back.** `ddcFailStreak` survives; `ddcAvailable` only latches false after three consecutive failed writes, and a failure inside the grace window just lets the pump take the next target. The notice logged at the third failure and the behaviour `docs/ddc-notes.md` documents are unchanged.

Two things `main` moved under the branch and this rebase had to integrate:

- **`DDCService.hold()`** (#110) hung off the global `ddcQueue` this PR removes. It is now `DDCOperationQueuePool.hold(timeout:onIdle:)`: it parks every per-display queue, including a display that gets its first operation mid-hold, resolves once already-queued work has drained, and each parked queue gives up after 15 s so a lost release cannot wedge DDC. Overlapping holds each release only the queues they parked. Three tests cover it.
- **Read quarantine and the slow-op logging** (#90, #110) assumed "accessed only on ddcQueue". That state moved behind a `readStateLock`, with the log calls outside the lock.

Also: the completion handler split into `ddcWriteSucceeded` / `ddcWriteFailed` (the merged version tripped SwiftLint's cyclomatic complexity cap), and `docs/ddc-notes.md` picked up the new pairing order, the per-display queues, the coalescing writer, and a failure-class row for a channel that blocks.

## Verification

`make check`: SwiftLint strict clean, 102 tests green, localization keys complete. All 8 commits compile individually, not just the tip.

Hardware, two AOC U32N10 on an M4 Max — the pathological pairing case, since both report vendor `0x05E3` product `0x3210` **serial 0**, so identity alone cannot tell them apart:

- **Pairing.** CoreDisplay puts display 2 at `dispext1@90000000` and display 5 at `dispext0@B0000000`; the two `DCPAVServiceProxy` channels sit under those same framebuffers. Crisp paired 2 → channel 1 and 5 → channel 0, matching macOS's own mapping. Without the location step, model matching would have handed channel 0 (physically display 5) to the first unused display in online order — display 2 — and crossed the two sliders.
- **Ground truth, read straight off the channels with `scripts/ddc-probe.swift`.** Setting display 2 to 70 moved channel 1 alone (channel 0 stayed at 44); setting display 5 to 20 moved channel 0 alone.
- **Per-display queues are visible in the log.** Both displays' `0x62` probes completed in the same millisecond on two threads, and their `0x10` reads 2 ms apart on two more. On one global queue those serialise.
- **Coalescing.** 40 rapid sets on display 2 sweeping 70 → 31 landed exactly 31 on the panel, and a set to display 5 issued mid-burst (23.8 ms round trip) landed its 60 — neither dropped nor queued behind the burst.
- No slow op over 500 ms, no failed write, no read quarantine, no gamma latch anywhere in the capture.
- Both displays restored to their original 48/44 afterwards, verified by direct channel read.

What I could not reproduce on this hardware: the 12 s wedged channel (both panels are answering cleanly today), the three-failure latch (needs a monitor that drops writes — the Q27G3XMN case), and a replug mid-drag. The last one is what `DDCTopologyChangeTests` covers instead.

Closes #72.


didriksg (changes requested):
This is the right fix for the right cause. One global DDC queue was always going to let a slow monitor hold the others hostage, and per-display queues plus location-based matching are what I'd have reached for too. The timing numbers from the two AOCs are appreciated; that's the kind of evidence I can't get here.

Two regressions I found reading the diff, both in the multi-display path this PR is about:

1. `DisplayManager.refreshDisplays` now calls `invalidateDDCTopology(for:)` with every online external display on every add, remove or move event, not just the display that changed. `invalidateDDCTopology` cancels the animator, drops `pendingDDCTarget`, bumps the topology generation and clears `ddcAvailable` for each of them. So: drag display A's slider to 80, and while the write waits out the 50 ms pacing floor, display B replugs (or you rearrange displays in System Settings, which is a `.movedFlag` with no ID change). A's pending target is gone, `pumpDDCWrite` sees a stale topology and returns, and the software fallback never runs either because it checks the same generation. The slider says 80, the monitor never got it. Same thing truncates a running `setBrightnessSmooth` fade from the brightness keys or a preset. Before this PR only the removed display's state was cleared. Could you scope the invalidation to the IDs that actually disappeared or changed identity or location?

2. `ddcFailStreak` is deleted, so a single failed coalesced write now sets `ddcAvailable = false` until reconnect. `docs/ddc-notes.md` documents the three-failure latch on purpose: the AOC Q27G3XMN drops about half of its DDC commands at random, and with this change it would fall to gamma mode on the first drag. `DDCService.writeAsync` retrying three times within 100 ms doesn't cover that, since the monitor ignores the retries too. Either keep a multi-cycle grace before latching, or make the case for the change in the PR so the notes can follow.

Two smaller things: please drop `docs/superpowers/plans/...` and `docs/superpowers/specs/...` (they're your planning notes, and already out of step with the code, for example the `removeQueue` that never got implemented) and the `.worktrees/` line in `.gitignore`.

With those in, I'll run it on my displays and merge. Thanks for digging into this one.

DEOWL-kan:
Thanks for the review — both regressions were real, and the second one especially: I had dropped the latch without registering that `docs/ddc-notes.md` documents it on purpose. Rebased onto `4275339`; the planning docs and the `.gitignore` line are off the branch.

**1. Scoped topology invalidation.** `DisplayManager` keeps a fingerprint per external ID — vendor/product/serial plus the CoreDisplay location — and only invalidates the IDs whose fingerprint changed or vanished. Your scenario now behaves: A's pending target survives B replugging, a display appearing, and a rearrangement in System Settings (a `.movedFlag` with no identity change invalidates nothing at all). The location is what makes this work for identical panels — two monitors that swap IDs both come out changed. The diff is a pure function, `DDCTopologyChange` in `Models/`, and `DDCTopologyChangeTests` covers your case directly.

**2. Three-failure latch restored.** `ddcFailStreak` stays, `ddcAvailable` only latches false on the third consecutive failure, and a failure inside the grace window just lets the pump take the next target. No change to the notice or to what the notes describe. You were right that `writeAsync`'s three retries inside 100 ms don't cover a monitor that ignores the retries too.

Two things `main` moved under the branch while it sat, so the rebase had to integrate them rather than just replay:

- **`hold()` from #110** hung off the global `ddcQueue` this PR deletes. It is now `DDCOperationQueuePool.hold(timeout:onIdle:)` — parks every per-display queue, including a display whose first op arrives mid-hold, resolves once already-queued work drains, 15 s per-queue give-up so a lost release can't wedge the session, and overlapping holds release only what they parked. Three tests. Worth a look, since it is the piece I rewrote rather than merged.
- **Read quarantine and the slow-op logging** (#90, #110) were documented as "accessed only on ddcQueue"; that state is behind a `readStateLock` now, log calls outside the lock. Related: the registry walk no longer test-reads every `DCPAVServiceProxy`, so the six-second engine hold your #110 note describes is gone from the walk itself — the 20 s pairing-miss cache now only saves the walk cost, which is still worth having.

Also split the write completion into `ddcWriteSucceeded` / `ddcWriteFailed` (merged, it tripped the cyclomatic complexity cap), and folded the new behaviour into `docs/ddc-notes.md`: pairing order, per-display queues, the coalescing writer, and a failure-class row for a channel that blocks. And you were right about `removeQueue` — it exists in one commit and gets removed two commits later, because a reused display ID has to land on the same queue; it is gone from the final state.

`make check` is clean: SwiftLint strict, 102 tests, localization keys. All 8 commits compile individually.

Re-ran it on the two AOC U32N10s, which turn out to be the pathological pairing case — both report vendor `0x05E3` product `0x3210` **serial 0**, so identity alone genuinely cannot separate them. CoreDisplay puts display 2 at `dispext1@90000000` and display 5 at `dispext0@B0000000`, the two `DCPAVServiceProxy` channels sit under those same framebuffers, and Crisp paired 2 → channel 1, 5 → channel 0. Without the location step, model matching hands channel 0 (physically display 5) to the first unused display in online order — display 2 — and the sliders come out crossed. Read straight off the channels with `ddc-probe`: setting display 2 to 70 moved channel 1 alone, setting display 5 to 20 moved channel 0 alone. The log shows both displays' `0x62` probes finishing in the same millisecond on two threads and their `0x10` reads 2 ms apart on two more, which is the isolation this PR is for. A 40-step burst on display 2 landed exactly its final value, and a set to display 5 issued mid-burst landed too, 23.8 ms round trip. Both panels put back to 48/44 afterwards.

What I still cannot show you from here: the 12 s wedged channel (both panels answer cleanly today), the three-failure latch (that needs a monitor that drops writes), and a replug mid-drag — that last one is what the new tests stand in for. Full details in the updated PR description.

didriksg (approved):
Both asks are in, and the rebase integrates the two things main moved under you. I read the whole diff and ran it here.

What I ran:

- Compile, strict lint, an x86_64 typecheck and `make test` on 1a802f3, all green.
- The location match: `IODisplayLocation` from `CoreDisplay_DisplayCreateInfoDictionary` is exactly the `IORegistryEntryGetPath` of the external's framebuffer node on this Mac, so the first strategy pairs on it. With one or two externals attached there is one external `DCPAVServiceProxy` per attached display, so dropping the test read adds no empty-port channels here.
- One external with a working channel: pairing, a single write, a 19-step burst with the final value landing and the gamma table where it should be (identity above the blend threshold, the blend below), a display sleep and wake with re-pairing and a write after it.
- One external whose reads fail and quarantine: the first write went through the software preview and settled to identity once the write was acknowledged.
- The hold, with two externals attached and a write to display A every 100 ms: Disconnect Display on B (disable reported after 151 ms), ten seconds, Reconnect (enable after 275 ms). The stream to A never failed, A stayed on DDC, B re-paired on the first op after the enable, and a WindowServer heartbeat at 50 ms saw no stall. The "waited for DDC to go idle" line did not print, so the drain stayed under 500 ms, which is what a fast channel should give.

Two small things, neither blocking:

- The pairing log says "by identity" for a location match. The label does not distinguish the new first strategy, so a capture cannot show that the location step did the work. A "by location" label would make the next report from two identical panels readable.
- `ddcWriteSucceeded` and `ddcWriteFailed` call `setSoftwareBrightness` while holding `ddcPumpLock`, and that is a WindowServer gamma write. Nothing deadlocks, the lock order is consistent everywhere I looked, but a slider event that wants the lock waits on the round trip. `applyLatestSoftwareBrightness` does it outside the lock; the same shape would do here.

Not covered here: one slow channel blocking another, which your two identical panels cover, and the three-failure latch, which needs a monitor that drops writes.

One observation, not on this PR: after a quarter of an hour of ten writes per second to A, its next reads failed six times in a row and quarantined. A real drag never sustains that, and the old single queue would interleave the same way.

Thanks for the careful rebase. The hold rewrite reads better than what it replaced.

didriksg:
Main moved under this one after the approval: #130 turned displayIdentity into a static helper shared with the luminance lookup, which is exactly the function this PR gives a location parameter. Rather than ask you for a fourth round over two mechanical hunks I merged main into the branch myself: your location parameter now sits on the shared static function with a nil default, the call site passes ioRegistryPath as before, and nothing else in the diff changed. Compile, strict lint, the x86_64 typecheck and the tests are green on the merged tree here. I also put your ACKNOWLEDGMENTS line on the branch so it lands with the PR. Merging once the build check is through.

PR #73, opened 2026-08-24
Add crispctl automation control plane

Juns-g:
## Status: closed reference; do not merge

The maintainer set a clear review and maintenance boundary. This large draft is retained only as technical and real-device-validation reference and is not a merge proposal.

The final archival branch head includes the locally verified follow-up commits through `80cd1b57db92eb4f0282be011fa7bb457eaa609e`. These commits preserve the completed batch-safety, framebuffer-identity, and fail-closed display-recovery experiments for future reference.

A later uncommitted experiment that attempted to settle disconnect timeouts was **not** included: independent review found that the WindowServer worker can continue after the caller times out, so an interim offline inventory cannot prove final success. Timeout therefore remains indeterminate and non-retry-safe in the archived head.

## Replacement delivery: minimal PR #75

The independently reviewable replacement is #75 and contains only the first brightness-control slice requested by the maintainer:

- list displays
- get brightness
- set brightness

The replacement deliberately excludes protocol versioning, batch/restore policy, Extra Brightness, and disconnect/reconnect.

## Historical evidence

The P0/P1 implementation and real-device findings in this PR and its comments remain reference material only. They do not establish validation for PR #75 or any future follow-up.


Juns-g:
P0 host-verification update (2026-08-25): Swift Package, 59 distribution tests, the independent-process Unix-socket round-trip, changed-file strict SwiftLint (25 files, 0 violations), and `git diff --check` are green. The repository-level `make test` failure in `DisplayManager.swift:113` reproduces unchanged at baseline `2975748` under the current Xcode beta. Host compilation did find two P0 blockers in `BrightnessBoostService`: the initial MonitorPanel ABI inspection used a Swift-unavailable `methodSignature(for:)` path, and `controlHeadroomSnapshot` missed a `return`. These are being fixed before any commit. `make loc-check` will be rerun after compilation. No real display write, signing, notarization, release, or Homebrew update has been performed.

Juns-g:
P0 host-verification follow-up (2026-08-25): the two P0 compile blockers reported above are fixed. `make compile` now succeeds, localization validation reports all 208 extracted keys present, Swift Package tests and all 59 distribution tests pass, the 25 changed Swift files remain strict-lint clean, and the default non-publishing `make build` dry-run produced a universal arm64+x86_64 app with an embedded universal `crispctl`. Strict ad-hoc signature verification and `hdiutil verify` pass; DMG SHA-256 is `c81aa31205f13f79ded2d6ec028fe83f303a4f8d64db89ac032dd8059a59c824`. Two independent fail-closed reviews are now running; no commit will be made unless both return without blockers. Repository-level `make test/check` remain baseline-blocked by unrelated existing diagnostics and generated `.build/out` lint input.

Juns-g:
Second fail-closed review update (2026-08-25): commit remains blocked. Reviewers found three logic gaps despite the green host build: (1) a final batch member could still be recorded successful after the absolute monotonic deadline, (2) Extra Brightness/HDR command settlement swallowed cancellation after mutation instead of propagating the indeterminate-write contract, and (3) Extra Brightness `off` was blocked after live eligibility collapsed, preventing persisted/active cleanup. Main review also confirmed the dispatcher capability guard must be fixed alongside the App host. A single Codex writer is adding deterministic RED→GREEN tests and minimal fixes. No commit/push or real display write has occurred.

Juns-g:
Third fail-closed review update (2026-08-25): protocol/deadline/cancellation review now passes, and the App/Core eligibility-loss cleanup implementation also passes. Commit remains blocked by one documentation mismatch: the canonical Agent Skill still tells agents to stop on every unsupported capability, contradicting the narrowly supported same-UUID Extra Brightness cleanup `off` path. The same single Codex writer is adding a focused RED→GREEN Distribution test and aligning only the canonical Skill plus crispctl docs. No implementation change, commit, push, or hardware write is being made in this step.

Juns-g:
Historical verification note for test artifact `1b7e2b5` (superseded in part by the correction below).

The package-level and distribution checks listed here passed locally, but the claimed independent-process Unix-socket round-trip was later found not reproducible at this commit because the script hard-coded an obsolete SwiftPM output path. See the correction in the next review comment: https://github.com/didriksg/Crisp/pull/73#issuecomment-5405355863

Real-device observations for the supplied `1b7e2b5` artifact, including the Extra Brightness async outcome issue, are recorded separately: https://github.com/didriksg/Crisp/pull/73#issuecomment-5408216795

Current PR-head and hosted-CI status are maintained in the PR description; this comment is retained only as historical evidence.

Juns-g:
Verification correction for the current PR head `1b7e2b5b2aa16b2cd9e0cdb35e3d9b281638ccd3`:

I reproduced the documented independent-process Unix-socket round-trip from that exact ref using a fresh scratch directory, and it currently fails before the fixture host starts. `scripts/test-crispctl-roundtrip.sh` hard-codes:
- `$SCRATCH/out/Products/Debug/crisp-control-test-host`
- `$SCRATCH/out/Products/Debug/crispctl`

But `swift build --scratch-path "$SCRATCH"` emits the binaries under the platform-triple directory (on my machine: `$SCRATCH/arm64-apple-macosx/debug/`). The concrete failure is at script line 27: `No such file or directory`.

This means the claim that the independent CLI-process ↔ Unix-socket-host round-trip passed is not reproducible from the submitted commit. Please resolve the binary directory dynamically (for example via `swift build --show-bin-path`), run this script in CI, then update the verification note once green.

The IPC/CLI architecture and safety semantics look strong; this is a verification-chain blocker, not a request to redesign the feature.

Juns-g:
Supplemental real-device validation on PR head 1b7e2b5 (Crisp 1.5.0 test artifact):

Validated successfully:
- Built-in display: brightness set/readback including >100% logical EDR brightness; 130% applied immediately and remained stable across immediate, 0.5 s, and 2 s reads.
- DELL U2723QE: DDC brightness write to 80% changed the physical display (manual visual confirmation). Its hardware brightness readback remains unavailable.
- LS27D80xU: software_gamma fallback changed the visible output at 80% and restored to 100% (manual visual confirmation). This is not a hardware OSD brightness guarantee.

Non-blocking but actionable issue:
`extra-brightness set builtin off` is asynchronously applied but reports a terminal failure too early. The immediate response was `ok: false` / `write_verification_failed`; immediate state was transitional (`enabled: true`, `persistedEnabled: false`), and it converged about 0.5 s later to `enabled: false`, `persistedEnabled: false`. Re-enabling succeeds immediately. Please either wait for convergence before returning a terminal result, or expose an explicit accepted/pending outcome plus a documented polling contract. This matters for automation, which could otherwise retry an operation already in flight.

Suggested docs/Skill clarification: distinguish (1) DDC control with verified hardware readback, (2) DDC write-only, and (3) software_gamma fallback. For the latter two, "read current value -> temporary change -> automatically restore" is not safe or guaranteed.

I did not test `set-all`: with one external display write-only and another gamma-only, a batch mutation cannot be safely restored to its pre-test hardware state. HDR has no writable target on this three-display setup.

Juns-g:
Scope correction: the real-device observations in my previous comment were performed against the locally supplied Crisp 1.5.0 test artifact built from `1b7e2b5`, not the current remote PR head `a47e637`. They remain valid reproduction evidence for that artifact, but current-head confirmation requires a freshly built/distributed artifact (or equivalent local build) from `a47e637`.

Juns-g:
The verified CI workflow update is now pushed in `f55f4cb` and adds the DistributionTests plus the headless `crispctl` Unix-socket round trip to CI. The exact new commands pass locally (65 Distribution tests and `CRISPCTL_HEADLESS_ROUNDTRIP_OK`). GitHub has held the fork workflow before creating any jobs, so could you please use **Approve and run** for this run? https://github.com/didriksg/Crisp/actions/runs/32871477850

The PR remains Draft; approving the run only lets the CI execute and does not mark it ready or merge it.

didriksg:
CI running is approved

Juns-g:
## Batch brightness: retain the strict default, add an explicit human override

Real-device validation showed that `brightness set-all 60` safely refuses before making any changes when an external display is DDC-writeable but has no readable brightness snapshot. That strict, automatically restorable mode is the right default for unattended automation.

However, `crispctl` is also an interactive human-directed CLI. An explicit user request should be able to perform a best-effort batch write when the user knowingly accepts that a particular display cannot be read back or automatically restored. The current all-or-nothing refusal makes that valid use case impossible.

Please keep the existing default and add a separate, deliberately named override mode (the exact flag/name is an implementation decision). It should:

- require an explicit acknowledgement of every write-only/unverifiable display;
- keep capability/range/identity preflight for every display;
- return per-display outcomes, distinguishing verified write, accepted-but-unverifiable write, failure, and not attempted;
- never describe the override result as automatically restorable or transactional;
- remain unavailable for display disconnect/reconnect, where the current fail-closed identity guard is still appropriate.

This is a product-completeness gap, not a request to weaken the safe default.

didriksg:
Thanks for the work you've put into this, and for the real-device notes in particular. The U2723QE and LS27D80xU runs are more careful validation than most PRs get.

I have to be straight about where this lands, though. It's 71 files and about 13,000 lines: a new Swift package, a versioned socket protocol, a 744-line display toggle service, a rewrite of a good part of BrightnessBoostService, and it grows with every update. I'm one person maintaining Crisp in my spare time. I can't review that with the care it needs, and every change I make to brightness or disconnect from now on would have to keep all of it working. So I won't merge this PR, and I'd rather you didn't spend more time growing it.

What I would take is a first slice small enough that I can read in an evening. Three commands: list displays, get brightness, set brightness, over a small local socket. No changes to the existing services beyond the hook they need, no protocol versioning, no batch-write or restore-snapshot policies yet. A few hundred lines, with tests for the pure parts in Models/. If that lands and people use it, disconnect and Extra Brightness can follow one at a time.

If that sounds reasonable, open it as a fresh PR against main and I'll review it properly. I'll keep this one open a little while as reference, then close it.

Juns-g:
That is reasonable. Thank you for the clear maintainer boundary and for considering the direction.

I will stop extending this PR and treat it as reference only. I will prepare a fresh, small PR against main with only local-socket display discovery plus brightness get/set, minimal hooks into the existing services, and focused tests for pure model logic.

I will keep protocol versioning, batch/restore behavior, Extra Brightness, and disconnect/reconnect out of that first slice. They can be proposed later as separate PRs only if the initial slice lands and proves useful.

For clarity: I also have additional local experimental changes that have not been committed or pushed. They are not part of this PR head, will not be added to this large draft, and will be evaluated only as separate, narrow follow-ups after the minimal first slice is reviewed.

Juns-g:
Closing as agreed with the maintainer. The final archival head is `80cd1b57db92eb4f0282be011fa7bb457eaa609e`; the unsafe timeout-settlement experiment was excluded. The small replacement is #75.

PR #75, opened 2026-08-27
Add minimal local brightness control socket

Juns-g:
## Scope

A small, unversioned, same-user Unix-socket control surface with exactly three commands:

```text
crispctl displays list
crispctl brightness get <display-id>
crispctl brightness set <display-id> <percent>
```

The complete diff is 12 paths and `+743/-4`: 461 production lines, 159 focused pure-model test lines, 17 lines of project wiring, a compact `+22/-4` documentation update across the README and English/Chinese site pages, and an 84-line minimal agent Skill.

The documentation explicitly describes this as source-build-only: the current public Crisp 1.5.0 DMG and Homebrew cask do not ship `crispctl`.

`skills/crispctl/SKILL.md` teaches agents only this three-command slice. It requires fresh numeric display discovery, explicit authorization before writes, exact accepted/queued/unverified response handling, separate app/model read-back, distinct human visual evidence, and no automatic retry. It deliberately excludes every later capability listed below.

## Set semantics

Brightness writes use Crisp's existing asynchronous service path. A successful set response therefore reports only:

```json
{"ok":true,"status":"accepted","execution":"queued","verification":"unverified"}
```

It does not claim hardware application, read-back, or automatic retry. The CLI validates this exact command-specific response before exiting successfully.

## Local-socket boundary

- socket mode `0600` in the per-user temporary directory;
- owned-socket checks and same-UID `getpeereid` checks on both sides;
- bounded LF framing and two-second I/O timeouts;
- no app auto-launch.

## Deliberately excluded

- protocol versioning;
- batch or `set-all` writes;
- restore snapshots / transactional policy;
- Extra Brightness, HDR, and display disconnect/reconnect;
- release/Homebrew packaging changes.

## Verification

- 79/79 XCTest cases passed, with 0 failures and 0 skips, verified from the generated `.xcresult`;
- the `crispctl` target builds with warnings as errors;
- strict SwiftLint, HTML parsing for all four edited site pages, bilingual claim checks, and `git diff --check` passed;
- the Skill frontmatter, nested Hermes metadata, exact three-command inventory, README link, unsupported-command exclusions, line count, and size checks passed;
- `make build` produced a universal `x86_64 + arm64` app/DMG dry run;
- GitHub CI passed on the implementation commit; the documentation/Skill-only head's CI run is currently `action_required`, awaiting upstream approval to run;
- usage, invalid-percent, unknown-display, and unavailable-socket paths return structured JSON with stable nonzero exits;
- independent security/protocol review: PASS;
- independent maintainer-scope/simplicity review: PASS.

Local Xcode 27 Beta promotes an unchanged `DisplayManager.swift:113` warning to an error under the repository's aggregate warnings-as-errors `make test` / `make check` commands. The exact same diagnostic was reproduced on `origin/main` at `aaf6b661eaab7cec1becec969cd8398f720027c6`, so it is a verified baseline/toolchain issue rather than a regression from this PR. With that warning non-fatal, the full 79-test suite passes; the changed `crispctl` target itself passes its strict warnings-as-errors build.

### On-device acceptance

The executable implementation at `9a068e4a2e550770173c56e2ec29d73ef9f7ab8b` was exercised using an ad-hoc local test build on macOS 27.0. The current head `dc0394e09a80aa07d8c0aa684b3165b00a202581` adds documentation and the minimal Skill only; its executable sources are unchanged.

#### Built-in display

- `displays list` discovered the built-in display as numeric ID `1`;
- `brightness get 1` returned its current logical brightness;
- one authorized write changed brightness from 100 to 90, and a separate read returned 90;
- one authorized restoration changed brightness from 90 to 100, and a separate read returned 100;
- both writes returned exactly `accepted / queued / unverified`, with no automatic retry;
- the user visually confirmed that the panel dimmed and then returned to its original brightness;
- graceful app termination removed the socket, the stopped CLI returned a structured transport failure, and reopening the same test build recreated the mode-0600 socket and restored successful reads.

#### External XG27UCS

- the external display was enumerated as numeric ID `2` alongside the built-in display;
- an isolated run started from 100 and issued exactly one authorized `brightness set 2 50`, with no restoration or retry during observation;
- `brightness get 2` returned 50 immediately and remained at 50 after one, five, and ten seconds;
- the user confirmed that the physical panel dimmed to approximately 50% and that the Crisp GUI also displayed 50;
- one final authorized restoration to 100 returned the accepted/queued/unverified contract, and independent reads immediately, after one second, and after five seconds all remained at 100;
- no write was retried.

An earlier one-off observation of a stale value after a rapid set/restore sequence was not reproduced and is not treated as a confirmed defect. One request issued immediately after an app launch hit the two-second receive timeout while the app and socket remained alive; later requests succeeded. This startup-window observation has not reproduced during the subsequent stable read/write runs.

The local test build was ad-hoc signed, not notarized, and not published.


Juns-g:
@didriksg I opened this fresh, deliberately small PR following the scope you outlined on #73: exactly `displays list`, `brightness get`, and `brightness set` over a same-user local socket, with pure-model tests and no protocol versioning, batch/restore policy, Extra Brightness, or disconnect work. I also added a compact README, English/Chinese site notes, and a minimal agent Skill; all of them are explicitly source-build-only and describe only these three commands.

For transparency about how it was developed: I directed Codex through Hermes to implement the changes, while I supervised the scope and required repeated unit tests, strict builds/lint, protocol and security review, packaging checks, and read-back verification. The real-device acceptance was supervised and visually confirmed by me personally on my own MacBook: the built-in display worked correctly, and a DDC-capable external XG27UCS was successfully discovered and changed between 100% and 50%, with matching `crispctl` read-back, Crisp GUI state, and visible brightness changes. I also personally confirmed restoration to the original state. The PR description records the exact evidence and the limits of what each check proves; in particular, queued/unverified responses are not presented as independent hardware read-back.

If it is useful while reviewing, please feel free to use AI-assisted code review as an additional way to inspect the small diff or challenge any part of the implementation; the tests and reproducible evidence are there so those findings can be checked rather than taken on trust. I have tried to validate this unusually carefully, both automatically and on real hardware. When you have a chance, could you please review it and approve it if it now matches the first slice you had in mind?

didriksg (changes requested):
This is the slice I asked for, thanks for cutting it down. The XG27UCS run with the 1/5/10 s read-backs is more care than most PRs bring. The socket code reads fine, and the flock ordering means a second instance can't unlink the live socket. Two fixes and two cuts, then I'll merge.

Fix:
- CI never compiles `Sources/crispctl/main.swift`; `make test` builds the Crisp scheme only. Add `crispctl: all` under `schemes.Crisp.build.targets` in project.yml so the test build covers it.
- `CrispControlModelTests.swift` still has the `#if canImport(CrispPureModels)` guard from #73. Nothing defines that module now, drop it.

Leave out of this PR:
- `skills/crispctl/SKILL.md` and the README link to it. I looked at what BetterDisplay and Lunar do: a few lines in the README, the reference in a wiki page or `--help`, and no agent skill file. The README section you wrote already says everything an agent needs (the three commands, Crisp must be running, JSON out, set is queued), so let's keep the CLI docs there. If the skill is useful for your own setup, keep it in your repo.
- The four `docs/*.html` edits. The site describes the app people download, and 1.5.0 doesn't have crispctl. I'll update the comparison table when it ships in a release.

One design point, your call but I lean toward changing it: the set reply carries `status/execution/verification` that can only ever hold one value each, and `classify` rejects any reply whose keys aren't exactly the expected set. With no protocol version, a crispctl someone built today breaks as soon as a later app adds one field to any reply. I'd make set return `{"ok":true}` and have the CLI accept any decodable reply with `ok`. The README already says a set is queued and not verified, which is where that belongs.

Small: the README should say how to build it (`xcodegen generate && xcodebuild -scheme crispctl -configuration Release`), and that a set counts as a manual change like the slider, so it clears the active preset. I'd cut the "no --json option" and "local test packages" sentences. I'll squash on merge.

didriksg (approved):
All points addressed, and the update checks out on a test merge with main: lint, tests, and a live list/get/set against the app. Merging, thanks.

didriksg:
Merged, and thank you for the work you put into this, @Juns-g. The 1/5/10 s read-backs on the XG27UCS, the flock and getpeereid hardening on the socket, and the clean rewrite after review all made it an easy merge. Your line in ACKNOWLEDGMENTS went in with #78.

Now that the slice is in, I'd be glad to see more of what you built in #73 come over the same way: one command group per PR, hooked into the services that already exist rather than reworking them. In the order I'd take them:

1. `displays list` growing a stable `uuid`, the current resolution, and whether brightness is DDC or software. Additive fields only; the CLI already tolerates unknown ones.
2. Extra Brightness: `brightness boost get/set <id>` on top of BrightnessBoostService as it is.
3. Disconnect and reconnect: `displays disconnect/reconnect <id>` through PhysicalDisplayToggleService, keeping its last-active-display guard. The recovery machinery from #73 stays out; the app already handles wake and replug.
4. Presets: `preset list` and `preset apply <name>`. Not in #73, but it's the command I'd use most myself.

Still out: protocol versioning, batch or transactional writes, packaging and release changes (shipping crispctl in the DMG is a call I'll make once it has earned it), and the skill file. #79 adds `crispctl help` as the reference for people and agents; each new command should extend that text and the README lines.

PR #81, opened 2026-08-28
crispctl: add display metadata

Juns-g:
## Scope

Extends the existing `CrispControlDisplay` response with additive metadata only:

- stable display `uuid`;
- current logical/pixel resolution, refresh rate, and HiDPI state when known;
- `brightnessBackend`: `builtin`, `ddc`, `software`, or `unknown`.

`displays list` is the intended consumer; `brightness get` receives the same fields naturally because it already returns the shared display DTO. No commands, arguments, socket behavior, exit codes, set semantics, or response-classification rules change.

The backend snapshot is read-only: it uses existing state under the existing lock, does no DDC I/O/probe/retry/write, and reports HDR software dimming as `software` before any prior DDC-success state.

This is independent of the unmerged #71 DDC-latency work and does not copy or depend on it.

## Verification

- `git diff --check`
- `xcodegen generate`
- `xcodebuild ... -scheme Crisp ... SWIFT_TREAT_WARNINGS_AS_ERRORS=NO test` — 81 tests, 0 failures, 0 skipped
- `xcodebuild ... -scheme crispctl -configuration Release ... CODE_SIGNING_ALLOWED=NO SWIFT_TREAT_WARNINGS_AS_ERRORS=YES build`
- `swiftlint lint --strict --quiet`

The strict warnings-as-errors Crisp wrapper is blocked on this machine by the existing Xcode 27 Beta `DisplayManager.swift:113` `#ImplicitStrongCapture` diagnostic; the identical failure was reproduced from unmodified current `origin/main`.

## Hardware validation

**Not performed.** No Crisp App launch, `crispctl` connection to a running app, or real display/hardware read or write was performed for this PR.

### DIFF
--- a/Crisp/Models/CrispControlModel.swift
+++ b/Crisp/Models/CrispControlModel.swift
@@ -4,11 +4,46 @@ enum CrispControlSocket {
     static let path = FileManager.default.temporaryDirectory
         .appendingPathComponent("crispctl.sock").path
 }
+struct CrispControlResolution: Codable, Equatable {
+    let logicalWidth: Int
+    let logicalHeight: Int
+    let pixelWidth: Int
+    let pixelHeight: Int
+    let refreshRate: Double
+    let isHiDPI: Bool
+}
+enum CrispControlBrightnessBackend: String, Codable, Equatable {
+    case builtin
+    case ddc
+    case software
+    case unknown
+}
 struct CrispControlDisplay: Codable, Equatable {
     let id: UInt32
     let name: String
     let brightness: Double
     let isBuiltin: Bool
+    let uuid: String?
+    let resolution: CrispControlResolution?
+    let brightnessBackend: CrispControlBrightnessBackend?
+
+    init(
+        id: UInt32,
+        name: String,
+        brightness: Double,
+        isBuiltin: Bool,
+        uuid: String? = nil,
+        resolution: CrispControlResolution? = nil,
+        brightnessBackend: CrispControlBrightnessBackend? = nil
+    ) {
+        self.id = id
+        self.name = name
+        self.brightness = brightness
+        self.isBuiltin = isBuiltin
+        self.uuid = uuid
+        self.resolution = resolution
+        self.brightnessBackend = brightnessBackend
+    }
 }
 struct CrispControlRequest: Codable, Equatable {
     enum Command: String, Codable {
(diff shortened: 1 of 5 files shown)


didriksg:
Saw this went to draft, so I have stayed out of the way. No rush at all, and ping me whenever you want eyes on it.

And don't worry about the hardware test. I have a few panels on the desk here, DDC ones included, so I'll run `displays list` across them and post what `brightnessBackend` and `resolution` actually come back with. Nothing you need to do, and it works on the draft as it stands.

From a read of the diff the shape looks right to me: additive optional fields, the backend snapshot taken under `ddcAvailableLock`, which already guards `hdrDimmedDisplays` as well as `ddcAvailable`, and no DDC I/O on that path. First impression rather than a proper review, but nothing jumped out.

Juns-g:
Thanks — I’ve finished the local read-only hardware pass and marked this ready for review.

Validated from the PR build at `24308f6`:

- `displays list` returned a non-empty `uuid`, complete `resolution`, and `brightnessBackend: "ddc"` for an external XG27UCS (`1920x1080` logical / `3840x2160` pixels, 144 Hz, HiDPI).
- The built-in display returned complete metadata with `brightnessBackend: "builtin"` (`1512x982` logical / `3024x1964` pixels, 120 Hz, HiDPI).
- `brightness get` using fresh runtime IDs returned the same `uuid`, `resolution`, and `brightnessBackend` as the list response for both displays.
- No `brightness set` or other display write was run.

The model tests cover missing resolution and the backend classifier, but this machine did not physically exercise reconnect stability, software fallback, HDR software-over-prior-DDC, or `unknown`/no-current-mode states.

When you run your panels, the most useful additional checks would be:

1. `displays list` across the DDC panels you already mentioned;
2. if convenient, compare the same panel’s UUID before and after a physical reconnect (the runtime numeric ID may change);
3. only if one of your setups naturally enters it, confirm software fallback / HDR software dimming reports `software` even after prior DDC availability.

No need to force an artificial `unknown` or missing-mode case. Ready for the proper review whenever convenient.

didriksg (approved):
Ran `displays list` against the three panels on my desk, from the PR build, raw off the socket:

- Built-in: `builtin`, 1352x878 logical / 2704x1756 px, 120 Hz, HiDPI, uuid present
- Dell U2412M (DisplayPort, portrait): `ddc`, 1200x1920, 60 Hz
- AOC Q27G3XMN (through a hub): `unknown`, 2560x1440, 165 Hz

The AOC is the case your hardware could not produce, and it came out right. It failed six DDC reads tonight and went into read quarantine, and it still reports `unknown` rather than `software`. That is correct: a read quarantine does not latch `ddcAvailable`, only the write path does (the three-failure latch in ddc-notes), so availability really is undetermined for it. Good to know the classifier holds there.

The rest I read off the diff: additive optional fields with a nil-defaulting init, the snapshot taken under `ddcAvailableLock`, the classifier pure and tested, and the legacy-decode test. Nothing to change.

PR #87, opened 2026-09-01
Avoid preset reads during SwiftUI layout

dboleslawski:
## Summary

- Cache display presets and the active preset index in `DisplayProfileController.reload()`.
- Render `DisplayPresetView` from the cached controller state and update that state after a successful selection.
- Avoid MonitorPanel KVC reads from SwiftUI view bodies, where nested screen-parameter notifications can re-enter AttributeGraph and trigger a precondition crash.

Split from #86 at reviewer request so the crash fix can land independently.

## Testing

- `make check`

### DIFF
--- a/Crisp/Views/DisplayDetailView.swift
+++ b/Crisp/Views/DisplayDetailView.swift
@@ -13,6 +13,8 @@ import SwiftUI
 final class DisplayProfileController: ObservableObject {
     let display: DisplayInfo
     @Published var presetName: String = ""
+    @Published var presets: [DisplayPresetService.Preset] = []
+    @Published var activePresetIndex: Int?
     @Published var activeProfileName: String = ""
 
     init(display: DisplayInfo) {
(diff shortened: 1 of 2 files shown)


didriksg:
Verified on my M4 Max MacBook Pro on 26.5.1: the Preset list on the built-in XDR panel renders all eleven entries from the cached snapshot, switching to HDR Video moves the checkmark and the row subtitle, and switching back works. CI, lint and compile are green, merging. One honest note: I couldn't reproduce the nested run loop. A run-loop observer around the MonitorPanel getters (`displays`, `hasPresets`, `presets`, `activePreset`) shows zero activity and 0 ms with the XDR panel online, so on this machine those reads never pump the loop. If you still have the crash log, I'd like to see it, mostly to learn which macOS version does this. Reading a private framework from inside a view body is wrong either way. Thanks for splitting it out.

PR #102, opened 2026-09-02
crispctl: add Extra Brightness controls

Juns-g:
## Summary

- add `crispctl brightness boost get <display>` and `crispctl brightness boost set <display> on|off`
- resolve `<display>` using the existing runtime-ID or stable-UUID selector behavior
- return the persisted Extra Brightness state and live eligibility from `get`
- expose each display's logical `maxBrightness` and preserve logical brightness values above 100 in `display list` / `brightness get`
- allow `brightness set <display> <percent>` above 100 only when Extra Brightness is enabled, currently eligible, and the requested value is within the same live ceiling used by the GUI
- refuse disabled, ineligible, unavailable-range, and over-maximum boosted writes instead of silently clamping them
- settle the short GUI range-expansion animation before an accepted boosted CLI write so the requested logical value is not transiently clipped
- keep the existing 0...100 behavior unchanged
- give only boost `set` a 5-second receive timeout because the existing service can wait while enabling external HDR; all other receive timeouts and all send timeouts remain 2 seconds

`off` remains available for a connected display even when it is currently ineligible, so persisted state can be cleaned up. Success reports that Crisp accepted the app-level logical operation; it is not a claim of independent panel, EDR headroom, luminance, or hardware verification. A transport timeout is not retried automatically—read the current state before deciding what to do next.

This intentionally does not add presets, display connect/disconnect controls, a protocol version, packaging changes, or a parallel Extra Brightness implementation.

## Verification

- `make check` — passed with stable Xcode (strict SwiftLint, **88/88 XCTest cases**, and all **209** localization keys)
- strict universal Release build of the `crispctl` scheme — passed for arm64 and x86_64
- added-lines security scan and `git diff --check` — passed
- independent specification/control-path review plus neutral adjudication — no blocking issue in this focused diff
- GitHub CI — passed: https://github.com/didriksg/Crisp/actions/runs/33792286952

## Manual hardware verification

- installed and launched the private ad-hoc test build from `cbbcd5c05186ad99cdf90a9710ca0b7b3285cc87` without replacing the stable app
- verified the built-in display was eligible with a live logical ceiling of 200
- set 120 and 180 successfully; both values read back exactly and the tester observed a clear brightness increase
- requested 201 and verified an over-maximum error with brightness remaining at 180
- disabled Extra Brightness and verified brightness / maximum collapsed to 100; a 120 request was rejected without mutation
- enabled Extra Brightness and immediately requested 180 with no inter-command delay; the value read back exactly at 180 and the tester observed the display dim to 100 then return to high brightness
- observed `NSScreen` current EDR headroom above SDR (`3.2`, potential `16.0`) during the boosted state; this is app/OS state plus visual confirmation, not an independent luminance measurement
- left the connected external display unchanged
- after acceptance, disabled Extra Brightness and explicitly restored the built-in display to 100; final read-back reported brightness 100, `maxBrightness` 100, and `enabled: false`

No GitHub Release was published.


didriksg (approved):
Ran cbbcd5c on the three displays here, the built-in and two externals, merged onto main at a585c95, straight through crispctl. The model checks land where the README says they do, and the readbacks are exact.

Built-in, plain range first: `set` by id and by uuid (upper, lower and mixed case) at 60, 25, 0 and 100, and at every step Crisp's readback and an independent `DisplayServicesGetBrightness` read of the panel agree exactly. Then boost: `boost get` reports eligible and disabled; `set 1 150` is refused while boost is off; `boost set 1 on` answers in 21 ms and `maxBrightness` goes to 200; 150, 199.9 and 200 are accepted and read back exactly; 200.1 and 300 are refused with the live maximum in the message; `boost set 1 off` collapses the range and 120 is refused again. At 150 the panel's own brightness stays at 100 and the screen's EDR headroom goes from 1.0 to 2.67, so the boosted value reaches the overlay and not only the number. External 1 (no headroom, no HDR): ineligible, `on` refused, `off` accepted, 101 refused. External 2: eligible through HDR, `boost set 2 on` took 2.7 s (the HDR switch), so the 5 s receive timeout on that command is needed rather than cautious; backend reads `software` while HDR is on, 150 reads back exactly, `off` returns the panel to SDR. Bad arguments (-5, nan, `maybe`, a missing selector) exit 2. Everything else answered in 20 to 50 ms.

One thing worth a line in the README, not a change: `boost set off` collapses the value over about half a second, and a read inside that window shows brightness above the maximum (147.7 against 100 at 0 s, 102.7 at 0.25 s, 100 from 0.5 s). Scripts that read straight after `off` should expect that.

Also the collapse lands on 100, not on the value the display had before boost was enabled (external 2 went 50, 150, 100). That is the existing service's behaviour and the GUI does the same, so nothing for this PR.

PR #104, opened 2026-09-02
Make a held-disconnected display discoverable (caption, menu bar mark, notice)

ncchen99:
## What & why

Follow-up to #93 / #101, and a question rather than a merge request: I'd like your read on whether you want any of this, and which parts.

Thank you for the review on #93 — the Reconnect race you caught was real, and the point about `reconcile` being where the choice becomes durable is what made the patch small. #99 and #100 were both better than what I would have sent.

**Stacked on #101.** The first two commits here are that PR; only the last three are new. Happy to rebase once #101 lands, or to close this and reopen it as separate PRs.

The problem is one #101 creates. A disconnected display is switched off at the window server: it shows no signal, System Settings does not list it, and replugging the cable changes nothing, because the window server holds that state and not the cable. While the choice only lasted until the display next enumerated, that was survivable. Now that it lasts, someone who comes back to a monitor they disconnected months ago gets a black screen, replugs the cable, power-cycles the monitor, and has every reason to conclude the hardware died. The choice was theirs, but nothing on the machine can tell them so.

Three independent commits, each of which stands alone — take any subset:

**1. A caption under the Disconnect row** (`65a5c07`). One line, at the moment the choice is made: "Stays disconnected until you reconnect it here." The cheapest of the three, and the one I would keep if you only wanted one.

**2. A dot on the menu bar icon while any display is held off** (`84e0463`). Opposite corner and a different colour from the Keep Awake one, so both can be read at once; the badge factory is now shared rather than duplicated. Deliberately small — it is not an alert, it is the thing a puzzled user clicks, and clicking it is what shows the Reconnect row.

**3. A notification with a Reconnect button** (`45d710e`). The one I am least sure you want, since it is a new framework and a new permission. Two things that I hope make it defensible: authorization is requested at the first banner rather than at launch, so no prompt appears for anyone who never reaches this state; and it tells without deciding — Crisp never takes the disconnect back on its own.

## Why it does not auto-reconnect on a replug

That was my first instinct and I think it is wrong. A cable arriving cannot be told apart from a dock re-enumerating everything on it (my second monitor is on the same dock, so every dock replug would resurrect it), nor from a DisplayPort monitor being switched off and on at the wall, where the link drops entirely and HDMI's does not — so the same rule would behave differently per connector, in a feature whose value is being predictable. And any signal that resurrects displays automatically can resurrect one that is not really there. So: tell, never decide.

## How tested

On a MacBook Pro (M5, macOS 26.6.1), Developer ID signed build, with an Acer E241Y on HDMI through a j5create dock, held disconnected by Crisp.

Finding the trigger took measuring, and the result was not what I assumed. Unplugging and replugging that cable, registry entry IDs beside the service events as they fired:

```
[03:24:15] EVENT terminate id=4294977625        <- cable out
[03:24:15] framebuffer 4294969518:-             ...same object, name cleared
[03:24:23] EVENT publish   id=4294978031        <- cable back in
[03:24:24] framebuffer 4294969518:E241Y G0      ...same object, name restored
```

Two things follow. The window server never moves: a disabled display does not leave `SLSGetDisplayList` and never re-enters the online list, so there is no reconfiguration callback and no CoreGraphics event to hang this on — which is exactly why the user sees nothing either. And `IOMobileFramebufferShim` is the wrong service to watch: its object outlives the cable and only its properties change, so it publishes nothing (I shipped that version to myself first and it stayed silent through four replugs). `DCPAVServiceProxy` is the one actually created and destroyed with the cable. It carries no EDID, so it is the trigger and the framebuffer is the identity; the name returns about 0.4s after the event, hence the delayed reads.

End to end, on the build with the corrected trigger:

```
[03:30:32.340] EVENT publish            <- cable back in
   03:30:32     notice posted for "E241Y G0"; the Reconnect button brings it back
[03:31:03] and [03:33:02] replugs       <- no further notice: the per-day limit holding
```

`make check` is green (SwiftLint strict, tests, localization keys; both new strings carry zh-Hans). `UserNotifications` autolinks — no build-script change needed, verified with `otool -L` on the swiftc path that `dev.sh` and `release.sh` use.

One note if the dock-switching rule from #92 ever lands: the notification's Reconnect should also tell that rule the user asked for the panel, the same way the menu's Reconnect does, or the rule will take it away again seconds later. I have that line locally but left it out here, since the service it calls does not exist on this branch.

## Checklist
- [x] Builds locally (`./scripts/release.sh`, Developer ID signed)
- [x] `Crisp/Resources/Localizable.xcstrings` updated for both new strings
- [ ] Screenshot — happy to add one for the caption and the badge if you want the change at all

### DIFF
--- a/Crisp/App/AppDelegate.swift
+++ b/Crisp/App/AppDelegate.swift
@@ -157,6 +157,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
 
         setupStartupBehavior()
         setupStatusItem()
+        DisconnectNoticeService.shared.start()
 
         // Re-anchor the open panel when screens change: switching the main
         // display re-origins global coordinates, which would otherwise leave
--- a/Crisp/App/AppDelegate.swift
+++ b/Crisp/App/AppDelegate.swift
@@ -282,9 +282,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
                 // Give WindowServer 2 seconds to stabilize after wake before
                 // touching display state.
                 try? await Task.sleep(nanoseconds: 2_000_000_000)
-                dm.refreshDisplays()
-                // Re-disconnect any physical displays macOS re-enabled on wake.
-                await PhysicalDisplayToggleService.shared.reapplyOnWake()
+                // The refresh also re-disconnects any display macOS re-enabled on wake:
+                // it runs reconcile, which puts a remembered disconnect back.
                 dm.refreshDisplays()
                 try? await Task.sleep(nanoseconds: 500_000_000)
                 // WindowServer keeps settling for several seconds after wake: ICC
--- a/Crisp/App/AppDelegate.swift
+++ b/Crisp/App/AppDelegate.swift
@@ -47,6 +47,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
     /// Drives the menu-bar Keep Awake indicator (keep-awake indicator).
     private var keepAwakeCancellable: AnyCancellable?
     private var keepAwakeBadge: NSView?
+    private var disconnectedCancellable: AnyCancellable?
+    private var disconnectedBadge: NSView?
     private var panel: MenuPanel?
     /// The panel is NEVER ordered out once warmed: taking the backdrop surface
     /// off screen makes WindowServer replay its materialize bloom (the growing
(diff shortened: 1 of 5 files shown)


didriksg:
Taking the caption, and only that. It says the cost at the moment of the choice, adds no control, and matches the subtitles the panel already uses.

Not the dot and not the notification, for now. Keep Awake's dot marks something you switch on for a while and would otherwise forget about; a remembered disconnect is a standing arrangement, and for the people who use it on purpose the dot would sit there every day marking a state they chose. The notification is a new framework, a new permission asked at the most confusing possible moment, and a trigger specific to the DCP path, all for a case nobody has filed yet. Your tell-never-decide reasoning is right, and I would rather leave the door open for the day someone reports the black monitor than add either now.

Once #101 lands, rebase this down to the caption commit and I will take it straight away. Or fold it into #101 if that is less work for you; either is fine.

didriksg:
#101 has landed and there's been no word here since, so I've taken the caption commit over as #126, as it is and with your author line on it, and I'm closing this one. The dot and the notification stay out, for the reasons above. Thanks for the caption, it's the right sentence in the right place.

Issue #112, opened 2026-09-03
Blackout rescue waits out WindowServer's phantom displays after an undock while asleep


didriksg:
### What happens

Undock a MacBook while it sleeps, with the built-in panel held disconnected through Crisp, then wake it. All screens stay dark for about sixteen seconds. Then the built-in comes back.

### Why

Measured by @ncchen99 on #92 with a sampler (online list, active list, `CGDisplayIsActive`, and the value `physicalActiveDisplayCount` would return, every 200 ms, with a heartbeat). At sleep the externals leave the active list and the count reads 0. It stays 0 through deep sleep and the DarkWake after the unplug. At full wake WindowServer re-enumerates and brings both externals back as new displays with their real EDID identities and `CGDisplayIsActive` true, while the dock is physically out. They stay for 12.3 s until WindowServer notices the link is gone. During that window the count is 2, so `restoreIfNoActiveDisplay` stands down at its count guard on every reconfiguration callback. When the phantoms leave, the rescue runs its four-second confirmation and re-enables the panel.

The shape rule from #106 does not apply. It rejects vendor 0, model 0 and codes wider than 16 bits, which is what a stale id turns into later. These entries carry real vendor and model codes. "An id never seen healthy" is not usable either, since a normal wake reassigns ids as well.

### What would fix it

A signal below WindowServer that says the link is down while CoreGraphics still says active, if the framebuffer node in the IORegistry reports one during those twelve seconds. I have no dock here. A capture of the same run with #111's line (`restore asked with 1 record(s): 2 active display(s), in flight false, standing down`) plus the framebuffer's registry state in that window is what this needs. If nothing there differs, sixteen seconds is the cost and this closes as documented.

Related: #91, #92, #111.

didriksg:
@ncchen99 I ran the undock-while-asleep case on my own setup tonight, minus the dock: one external on a direct USB-C cable, Mac asleep, cable pulled, then woken. I wanted to know what below WindowServer changes at the moment the cable goes, and there is one thing that does.

[...]

PR #132, opened 2026-09-07
Re-read the count before the blackout rescue commits

### LINKED ISSUE #117: Blackout rescue's settle check can sample a stale display count and re-enable a display on a desk that is already back
# The blackout rescue's settle check can sample a stale display count and re-enable a display on a desk that is already back

Companion to #112, and the opposite direction. #112 is `physicalActiveDisplayCount`
reading too **high** — phantoms counted as active, so `restoreIfNoActiveDisplay` stands
down on a desk that really is dark. This is the same guard reading too **low**: a display
that is already back in the active list is not yet visible to the app, so the rescue fires
on a desk that is already recovering and re-enables a display the user deliberately
disconnected.

No cable is pulled anywhere in this report. The dock stays plugged in the whole time.
This is an ordinary wake.

## Setup

MacBook Pro Mac17,2 (M5), macOS 26.6.1. BenQ GW2780 (`09D1/78E6`) on the DisplayPort of a
j5create JCD552 dock, Acer E241Y G0 (`0472/0E2B`) on the dock's HDMI, built-in panel
(`0610/A05C`). Built-in and Acer both held disconnected, so the BenQ is the only screen.

Build is **main at 4275339, unmodified, in a scratch bundle**, the only Crisp instance
running: a clean worktree at that commit, `git status` empty, compiled with `dev.sh` into a
scratch bundle. None of the branch work I have open is in it (no `AutoDisplaySwitchService`,
no `reapplyRemembered`, no #104). The event that started this was on a branch of mine, so I
re-ran it on clean main before filing and it reproduced there.

## What happens

Sleep, wait 20-30 s, wake with a keypress. The BenQ leaves the active list for ~2 s while
its link retrains and comes back on its own. Sometimes, in that window, the rescue fires:

```
23:15:24.012  no active display, restoring from 2 record(s)
23:15:24.012  reconnect requested: 37D8832A-2D66-02CA-B9F7-8F30A301B230 id 1
23:15:24.450  enable 1: reported success after 438 ms
```

The built-in comes on and stays on, and `reconnect()` has dropped its record, so the
disconnect the user asked for is gone for good. Nothing on main puts it back.

## The measurement

`settle-probe.swift` (attached) samples `physicalActiveDisplayCount`'s own arithmetic every
20 ms from a separate process. `phantom-probe.swift`'s 200 ms is the right resolution for
#112's 12-104 s phantom but cannot see this one; the margins here are tens of milliseconds.

Six wake cycles on clean main produced an empty-desk window. One fired. The single thing
that separates them is whether a reconfiguration callback reached Crisp between the BenQ
actually returning and the settle check reading the count:

| cycle | probe sees BenQ back | Crisp's next callback | settle check | count read | outcome |
|---|---|---|---|---|---|
| 23:13 | 23:13:23.336 | 23:13:23.655 (**309 ms before**) | 23:13:23.964 | 1 | stands down |
| 23:14 | 23:14:16.074 | 23:14:16.372 (**298 ms before**) | 23:14:16.543 | 1 | stands down |
| 23:15 | 23:15:23.728 | 23:15:24.049 (**37 ms after**) | 23:15:24.012 | 0 | **fires** |

At 23:15:24.012 the BenQ had been back in the active list for **284 ms** — a separate
process polling the same call saw it at 23:15:23.728 with its EDID intact (`2[9d1/78e6]`).
Crisp read 0 anyway, and read 1 thirty-seven milliseconds later, inside a callback.

The original event, on my branch, has the same signature to the millisecond:

```
22:40:57.644  WindowServer: Display 2 setEnabled:1      (the BenQ is back)
22:40:57.934  Crisp: no active display, restoring       (count read as 0)
22:40:57.971  Crisp: restore asked ... 1 active display(s)   ← 37 ms later
```

Both firings lost the race by exactly 37 ms.

## Why the settle delay does not cover it

The 2 s window is not the operative number. `Task.sleep(2_000_000_000)` on a wake overran
by **418-1254 ms** across these runs, so the effective deadline is 2.4-2.7 s and varies per
wake. Two cycles had empty-desk windows *longer* than 2 s (2.253 s, 2.244 s) and correctly
stood down, because the display returned inside the overrun. Lengthening the settle delay
therefore does not fix this: the check would still be one sample, and one sample can be
stale whenever it lands.

Nor is it about sleep depth. The dark window is 1.26-2.25 s whether the machine slept 11
seconds or 2 hours 42 minutes (2.152 s vs 2.149 s on the pair that started this).

## Suggested fix

Re-poll instead of sampling once. The settle Task currently does:

```swift
try? await Task.sleep(nanoseconds: 2_000_000_000)
let settled = self.physicalActiveDisplayCount()
guard settled == 0 else { ...stand down... }
```

A short poll — say ten reads at 100 ms — before committing to the restore would have stood
down in both firings, which needed only 37 ms more. It costs nothing on the path the
rescue exists for: a genuinely dark desk stays dark for all ten reads.

This is a different fix from the one #112 needs. #112 needs a better *predicate* (a
transport node behind the display); this needs the count to be read more than once. Neither
covers the other.

## What I could not determine

- **Why the count is stale.** The measurements say the app's reading trails the true state
  until a callback arrives; they do not prove the mechanism. I tried to pin it with a probe
  that registers `CGDisplayRegisterReconfigurationCallback` and polls from a worker thread
  in one process (`cache-probe.swift`), but a command-line tool is never delivered the
  callback, so there was nothing to compare — the polling side tracked the truth live.
  Settling this properly needs the instrumentation inside a real app, i.e. inside Crisp.
- **Rate.** One firing in six empty-desk windows on main, plus the original. I have not
  characterised what makes a given wake fall on one side.

Ruled out along the way, in case they save anyone else the detour:

- **How the machine is woken.** The firing cycle was woken with a single keypress and the
  two stand-downs either side of it with a double tap on the space bar, which looked like a
  correlation worth chasing. It is not one: pmset reports an identical wake source for every
  cycle (`USB2_wake SMC.OutboxNotEmpty nub-spmi-a0.0x25/UserActivity Assertion`), and a
  further single-keypress wake stood down normally.
- **Sleep duration.** The empty-desk window is 1.26-2.25 s whether the machine slept 11
  seconds or 2 hours 42 minutes.
- **The length of the settle delay.** See above: two cycles exceeded 2 s and stood down
  correctly.

One number in this report — the extra single-keypress wake under "Ruled out" — is from my
own build rather than the scratch one. Everything in the tables is from main.

## Files

Held rather than pasted, the way we did it on #112 — say which and I will paste them in
full or attach:

- `settle-probe.swift` — the 20 ms probe (happy to send it as a PR against `scripts/`
  alongside `phantom-probe.swift` if you want it in the tree)
- the six cycles' probe output, and the seventh
- `subsystem == "com.crisp.app"` category `display` for the whole 23:00-23:16 window
- WindowServer's `setEnabled` timeline for the same cycles
- pmset sleep/wake transitions
- `cache-probe.swift` and its output — the inconclusive attempt at the mechanism

ncchen99:
Closes #117.

`restoreIfNoActiveDisplay` waits 2 s and then decides on **one** read of
`physicalActiveDisplayCount`. Re-read up to ten times at 100 ms instead, and stand down on
the first non-zero.

## Why one sample is not enough

The count comes from this process's own copy of the display configuration, and that copy is
only as fresh as the last reconfiguration callback delivered to it. On #117 the external
was back in the active list for **284 ms**, visible to a separate process polling the same
call, while this read still returned 0; the callback that would have said so arrived 37 ms
after the decision. Three firings across two builds lost by 37 ms, 37 ms and 158 ms.

A longer settle does not fix it. `Task.sleep(2s)` on a wake already overruns by 418-1254 ms
— two measured cycles had empty-desk windows *longer* than 2 s and stood down correctly
because the display returned inside the overrun. One sample is stale whenever it lands.

## The bound is the point

Ten reads, not a watch. During a display sleep `CGGetActiveDisplayList` reports zero
displays for as long as the screens are down — 2 min 57 s of it measured on #112 — with
nothing wrong and nothing to rescue. The only thing keeping this guard out of that state is
that it runs from reconfiguration callbacks, and none arrive while the displays are asleep.
An unbounded poll here would walk straight into it. (I have that mistake in a branch of my
own: a 1 Hz blackout watch, caught rescuing 13 s into a display sleep with `enable 1` then
failing at exactly 10000 ms.)

## Measured, on this branch

Eight display-sleep/wake cycles, clean main plus this commit, scratch bundle, the settle
task armed in all eight:

```
13:07:28.444  restore settled: 1 active display(s) after 2 s and 0 re-poll(s), standing down
13:08:04.745  restore settled: 1 active display(s) after 2 s and 2 re-poll(s), standing down
13:08:40.932  restore settled: 1 active display(s) after 2 s and 2 re-poll(s), standing down
13:09:17.186  restore settled: 1 active display(s) after 2 s and 2 re-poll(s), standing down
13:09:53.380  restore settled: 1 active display(s) after 2 s and 1 re-poll(s), standing down
13:10:29.644  restore settled: 1 active display(s) after 2 s and 2 re-poll(s), standing down
13:11:05.899  restore settled: 1 active display(s) after 2 s and 2 re-poll(s), standing down
13:11:42.150  restore settled: 1 active display(s) after 2 s and 2 re-poll(s), standing down
```

**Seven of the eight needed a re-poll**, i.e. the first read after the 2 s sleep still said
zero and one or two further reads 100 ms apart found the display. Those seven are exactly
the decision the old code would have made on the stale sample, and it would have restored.
One or two re-polls is 100-200 ms; the ten-read bound leaves nine tenths of a second of
headroom over the worst margin measured.

The one cost is on the path the rescue exists for: a genuinely dark desk now pays up to
1 s more before the built-in comes back, on top of the 2 s that was already there.

## Also in here

`scripts/settle-probe.swift`, next to `phantom-probe.swift`. It samples the same arithmetic
every 20 ms from outside the app and prints how long the desk was actually empty and which
side of the deadline the return landed on. `phantom-probe.swift`'s 200 ms is right for a
phantom that lives 12-104 s and cannot resolve a 37 ms margin.

`make check` green. Not driven: the genuinely-dark path — nothing here was unplugged, and
#112 is the run that exercises it.

### DIFF
--- a/Crisp/Services/PhysicalDisplayToggleService.swift
+++ b/Crisp/Services/PhysicalDisplayToggleService.swift
@@ -846,6 +846,15 @@ final class PhysicalDisplayToggleService: ObservableObject {
         }
     }
 
+    /// How many times the settle check re-reads the count before committing to a restore,
+    /// and how long it waits between reads. The count is per-process and refreshed by
+    /// reconfiguration callback delivery, so the read taken the instant the timer expires
+    /// can trail the truth (issue #117). Ten reads at 100 ms covers the margins measured
+    /// there without becoming a standing watch -- see the comment in the settle Task for
+    /// why the bound matters.
+    private static let settleRepolls = 10
+    private static let settleRepollInterval: UInt64 = 100_000_000
+
     /// Guards against overlapping restore attempts from reconfiguration-callback bursts.
     private var restoreInFlight = false
 
@@ -871,9 +880,35 @@ final class PhysicalDisplayToggleService: ObservableObject {
             try? await Task.sleep(nanoseconds: 2_000_000_000)
             guard let self else { return }
             defer { self.restoreInFlight = false }
-            let settled = self.physicalActiveDisplayCount()
+            // One sample here is not enough. The count comes from this process's own copy
+            // of the display configuration, and that copy is only as fresh as the last
+            // reconfiguration callback delivered to it: on #117 the external was back in
+            // the active list for 284 ms, visible to any other process, while this read
+            // still returned 0 and the callback that would have said so arrived 37 ms
+            // after the decision. Measured three times on two builds, the margins were
+            // 37 ms, 37 ms and 158 ms, so a few hundred milliseconds of re-reading covers
+            // it where a longer sleep does not -- a single sample is stale whenever it
+            // lands, and Task.sleep above already overruns by 418-1254 ms on a wake.
+            //
+            // Bounded to settleRepolls deliberately, rather than made a watch. The
+            // margins being answered are 37-158 ms, so ten reads already carry an order
+            // of magnitude of headroom, and what the count does during a display sleep is
+            // not something to lean on: measured on this desk it stayed at 1 through a
+            // 35 s sleep with Crisp running, and read 0 for 2 min 57 s on the same desk
+            // with Crisp quit. A poll outliving the settle window could therefore sit on
+            // a zero-count desk that is only asleep. What keeps this guard clear of that
+            // state is that it runs from reconfiguration callbacks, and none arrive while
+            // the displays are down.
+            var settled = self.physicalActiveDisplayCount()
+            var repolls = 0
+            while settled == 0, repolls < Self.settleRepolls {
+                try? await Task.sleep(nanoseconds: Self.settleRepollInterval)
+                guard !Task.isCancelled else { return }
+                settled = self.physicalActiveDisplayCount()
+                repolls += 1
+            }
             guard settled == 0 else {
-                Self.log.notice("restore settled: \(settled, privacy: .public) active display(s) after 2 s, standing down")
+                Self.log.notice("restore settled: \(settled, privacy: .public) active display(s) after 2 s and \(repolls, privacy: .public) re-poll(s), standing down")
                 return
             }
             // This path acts with every screen black, so without a line here a capture
--- a/scripts/settle-probe.swift
+++ b/scripts/settle-probe.swift
@@ -1,0 +1,121 @@
+// Settle-window probe: does the display come back before or after
+// restoreIfNoActiveDisplay's 2 s settle check, and by how much.
+//
+// Run: swift scripts/settle-probe.swift     (Ctrl-C to stop)
+//
+// phantom-probe.swift samples every 200 ms, which is the right resolution for a
+// phantom that lives for 12-104 s. It cannot resolve this one: the margin between
+// the display returning and the settle check firing is tens of milliseconds. This
+// samples every 20 ms and reports the zero-window directly.
+//
+// ACTIVE is physicalActiveDisplayCount's own arithmetic: CGGetActiveDisplayList
+// filtered by the 16-bit EDID shape check from #106. (The VirtualDisplayService
+// exclusion is not reproducible outside the app; with no Crisp virtual display on
+// the desk the two agree.) ONLINE is phantom-probe's form -- CGGetOnlineDisplayList
+// filtered by CGDisplayIsActive -- kept so the two can be cross-checked the way
+// they were on #112.
+//
+// On every 1+ -> 0 transition the probe opens a window and on the way back out
+// prints how long the desk was actually empty. The nominal 2.000 s in the verdict is
+// the optimistic end: Crisp's check is Task.sleep(2s) plus scheduling, which on a
+// wake was measured landing 418-1254 ms late, so a window "over" 2.000 s can still
+// stand down. Read it with the app's own `restore settled` / `no active display`
+// line, which is what the check actually decided.
+//
+// It prints on any change to the per-display detail too, not just the counts. A
+// display can enter CGGetActiveDisplayList before its EDID is populated, and #106's
+// shape filter reads vendor/model 0 as "no panel" -- so that window leaves the count
+// at 0 and is invisible to a counts-only probe. Lines showing a real display id
+// marked `nopanel` are that state.
+import CoreGraphics
+import Foundation
+
+let fmt = DateFormatter()
+fmt.dateFormat = "HH:mm:ss.SSS"
+
+/// physicalActiveDisplayCount's arithmetic, minus the VirtualDisplayService
+/// exclusion, which cannot be reached from outside the app.
+func activeCount() -> (n: Int, detail: String) {
+    var count: UInt32 = 0
+    guard CGGetActiveDisplayList(0, nil, &count) == .success, count > 0 else { return (0, "") }
+    var ids = [CGDirectDisplayID](repeating: 0, count: Int(count))
+    guard CGGetActiveDisplayList(count, &ids, &count) == .success else { return (0, "") }
+    var n = 0
+    var parts: [String] = []
+    for id in ids.prefix(Int(count)) {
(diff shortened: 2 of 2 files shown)


ncchen99:
The genuinely-dark path is driven now, which was the "not driven" line above. It came up
while validating #133, since that is the sequence that produces a real blackout: built-in
held disconnected, sleep, dock pulled while asleep, woken with the built-in keyboard.
Both branches merged on main, scratch bundle.

The rescue fired and the desk came back in 4.2 s, and the cost of the poll on that path is
exactly what the bound predicts:

```
13:40:58.993  probe: the phantom appears, nothing viewable
13:41:02.362  no active display, restoring from 2 record(s)
13:41:03.169  enable 1: reported success after 806 ms
```

3.369 s from dark to the decision: the 2 s sleep, ~0.37 s of overrun on it, and the full
1 s of re-polls, since `settled` stayed 0 for all ten reads with the dock out of the
machine. That is the worst case by construction — a desk that is really dark pays every
re-poll — and it is the second before the panel lights that I described as the cost.

No `restore settled` line in the window: the poll never stood the rescue down on a desk
that had nothing on it.

Together with the eight display-sleep cycles in the description, where seven needed a
re-poll and stood down, both directions are now measured: it does not fire on a desk that
is coming back, and it still fires on one that is not.

ncchen99:
Correction to something I said here and on #112, before it gets built on.

I wrote that during a display sleep `CGGetActiveDisplayList` reports zero displays for as
long as the screens are down, "measured at 2 min 57 s", and used it as the reason the poll
has to be bounded. That measurement is real but I generalised it past what it supports: it
was taken with **Crisp not running**. With Crisp running, the same desk behaves the other
way.

Four 35-40 s display sleeps, one desk, the only variable being whether the app is up:

```
Crisp quit      13:49:11  COUNTED=0  online[2[9d1/78e6 asleep]]  active[]        (held 40 s)
Crisp running   13:50:51  COUNTED=1  online[34[9d1/78e6]]        active[34[...]] (held 35 s)
Crisp running   13:51:41  COUNTED=1  online[36[9d1/78e6]]        active[36[...]] (held 35 s)
```

With the app up the display keeps its place in the active list, with real EDID, through the
whole sleep, and the count reads 1. With the app quit the active list empties and the
online list keeps the display flagged `asleep`. Both reproduce. Since the guard runs inside
the app, the case that matters is the one where the count did **not** go to zero, so the
hazard I described is not established for the situation it was describing.

I have not chased why the app's presence changes it, and I would not guess in a code
comment.

The bound stays, with the reasoning corrected — commit amended, `make check` still green.
It now says: ten reads cover the measured 37-158 ms margins an order of magnitude over, and
the count's behaviour during a display sleep is not something to lean on in either
direction, having read 1 and 0 on the same desk under the only variable I changed. That
makes the bound cheap insurance rather than a proven necessity, which is what the evidence
actually supports.

Nothing about the fix itself changes: same code, same measurements in the description, same
result on the dark path above.

didriksg (approved):
Approving on the measurements, which cover both directions: seven of eight sleep cycles needed a re-poll and stood down, and the real undock still fired. What I like about this one is that it adds no new signal, only reads the one we have more than once, and the bound keeps it from turning into a watch. The correction you posted about the count during a display sleep is the kind of note that saves the next person a week; thanks for leaving it in the comment. Merging as is.

Issue #91, opened 2026-09-02
restoreIfNoActiveDisplay stops at the first record that reports success, which need not be a viewable screen


ncchen99:
Filing this as a code-level observation rather than a bug report, because **I could not reproduce
a user-visible failure on 1.5.0** — I tried and macOS got there first (more below). Three things I
measured while working on my own display setup suggest the check is unsound anyway, and you are
in a much better position than me to judge whether it is worth hardening. Close it freely if not.

### The check

`restoreIfNoActiveDisplay` returns on the first record whose `reconnect` reports success:

```swift
for (record, _) in candidates {
    if case .success = await self.reconnect(uuid: record.uuid) { return }
}
```

A `.success` there means `SLSConfigureDisplayEnabled` plus `CGCompleteDisplayConfiguration`
returned `.success` for a display ID. The file already documents, twice and from hard experience,
that this is not proof of anything — `verifyBackOnline`'s doc comment says a successful
transaction "is NOT proof of recovery", and `softReconnect` deliberately ignores the result and
waits for enumeration instead. The restore path is the one place that still trusts it, and it is
the path that runs when there is no screen left to notice with.

### What I measured

**1. The candidate ordering has nothing to work with in that state.** Once the last real display
leaves, `SLSGetDisplayList` collapses to the placeholder display alone. On this machine, three
entries before and one after. So `resolveCurrentID` returns nil for every record, each falls back
to its stale `record.displayID`, and the built-in-first sort is asking `CGDisplayIsBuiltin` about
IDs the window server has already forgotten — which answers with garbage. Built-in-first is not
actually guaranteed at the moment it matters most.

**2. A display coming back online can be a monitor that is switched off.** I have a second monitor
plugged into my dock over HDMI with its power button off. It is enumerated as a completely
ordinary display, and I probed it side by side against the monitor I was actually looking at:

| | BenQ GW2780 (real) | Acer E241Y (powered off) |
|---|---|---|
| active / online / asleep | 1 / 1 / 0 | 1 / 1 / 0 |
| EDID vendor / product | 0x09D1 / 0x78E6 | 0x0472 / 0x0E2B |
| mode count | 174 | 163 |
| gamma table capacity | 1024 | 1024 |
| `NSScreen.localizedName` | BenQ GW2780 | E241Y G0 |

Nothing separates them. So "a display is active again" cannot mean "the user can see something",
for this loop or for anything downstream of it.

**3. During a blackout on this machine, that is exactly what came back.** With my only live
[...]

PR #133, opened 2026-09-07
Cap the active display count at what the ports can be carrying

### LINKED ISSUE #112: Blackout rescue waits out WindowServer's phantom displays after an undock while asleep

ncchen99:
Closes #112. Stacks with #132; they touch the same guard from opposite sides — that one
makes the count read more than once, this one makes it read the right thing.

Count the DisplayPort and Thunderbolt transport nodes with hot-plug detect asserted, and
cap the external display count at that.

## Why hpd and not node presence

From the #112 capture: this desk's empty `HDMI3` port keeps its node in **every sample of
every run**, reading `hpd=Unknown sink=0`, so a phantom could hide behind an empty socket.
The two USB-C nodes read `hpd=High sink=1` while their displays were up and were gone
6.6 s before the phantoms appeared. Hot-plug detect is the half that tracks the cable.

## The precondition you asked for

> that hpd stays High on your dock's ports through display sleep, since if it drops there
> the rule would call a sleeping display a phantom and the rescue would fire on a desk
> about to come back

It holds. 1893 samples across two runs — one with Crisp not running at all, both externals
active, a single 3-minute display sleep; one on clean main with the reported arrangement
and four 32 s sleeps. `predicate-probe.swift` computed this rule on every 200 ms sample and
flagged `effective < actual` and `effective == 0 while actual > 0`. **Neither ever
occurred.** The port carrying an active display never left `hpd=High sink=1`, through the
display sleeps and through every wake blackout window. Full numbers on #112.

One caveat, in the code comment too: hpd does go `Low sink=0` for 400-600 ms on a port
whose display *Crisp has just switched off* — two of those blips landed 20 ms and 151 ms
after a `Display 3 setEnabled:0` in WindowServer. A live display's port never did it, and
the cap is an upper bound so a low read on a switched-off display's port cannot remove an
active one. Worth knowing before anyone samples this inside their own toggle.

## nil, not 0

`liveDisplayPortCount()` returns nil when the machine exposes no transport nodes of either
class. That is "the signal is not available here", not "nothing is plugged in", and capping
on it would black out a desk this rule has never seen — I have no way to check every Mac
that is not this one. A machine that *does* expose the nodes and reports none asserted is
exactly the #112 state, and 0 is the right answer there.

## Tests

The rule is a plain function in `Crisp/Models/PhantomPortCap.swift`, added to the test
target the way the other model types are, with six cases: the phantom state, the built-in
never being capped, an ordinary docked desk, more ports than displays, one of two cables
gone, and the unavailable-signal passthrough.

`make check` green.

## Not driven

The phantom state itself. Reproducing it needs a dock pulled during sleep, and I did not
want to hand you a fix validated only against the arithmetic — but the capture on #112 is
from exactly that sequence on this hardware, and the rule is read off it. I can run the
undock-while-asleep sequence against this branch and post the probe output if you would
rather have that before merging; say the word and it is one evening.

### DIFF
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -39,7 +39,7 @@ SOFTWARE.
 - [@crayonlu](https://github.com/crayonlu) ([#53](https://github.com/didriksg/Crisp/pull/53))
 - [@Juns-g](https://github.com/Juns-g) ([#75](https://github.com/didriksg/Crisp/pull/75), [#81](https://github.com/didriksg/Crisp/pull/81), [#102](https://github.com/didriksg/Crisp/pull/102), [#116](https://github.com/didriksg/Crisp/pull/116))
 - [@dboleslawski](https://github.com/dboleslawski) ([#87](https://github.com/didriksg/Crisp/pull/87))
-- [@ncchen99](https://github.com/ncchen99) ([#101](https://github.com/didriksg/Crisp/pull/101), [#104](https://github.com/didriksg/Crisp/pull/104))
+- [@ncchen99](https://github.com/ncchen99) ([#101](https://github.com/didriksg/Crisp/pull/101), [#104](https://github.com/didriksg/Crisp/pull/104), [#132](https://github.com/didriksg/Crisp/pull/132), [#133](https://github.com/didriksg/Crisp/pull/133))
 - [@celsinho17](https://github.com/celsinho17) ([#97](https://github.com/didriksg/Crisp/pull/97))
 - [@alexlee2046](https://github.com/alexlee2046) ([#83](https://github.com/didriksg/Crisp/pull/83))
 - [@DEOWL-kan](https://github.com/DEOWL-kan) ([#71](https://github.com/didriksg/Crisp/pull/71))
--- a/Crisp/Models/PhantomPortCap.swift
+++ b/Crisp/Models/PhantomPortCap.swift
@@ -1,0 +1,30 @@
+import Foundation
+
+/// How many active displays have something behind them, given what the ports say.
+///
+/// `physicalActiveDisplayCount`'s shape filter rejects an entry with no panel behind it,
+/// but it cannot reject the third kind, from #112: after an undock while asleep,
+/// WindowServer re-enumerates the absent externals at the full wake with their EDID
+/// identities intact, so they are active, 16-bit and completely real-looking, and stay
+/// that way until the dock goes back in. What actually left with the cable is the port's
+/// transport node, so the count of externals is capped at the number of ports that can be
+/// carrying one.
+enum PhantomPortCap {
+    /// - Parameters:
+    ///   - builtin: active displays that are the built-in panel. Never capped: the panel
+    ///     is not on a port, so what the ports say has nothing to do with it.
+    ///   - external: active displays that arrived through a port.
+    ///   - portCap: ports with a DisplayPort or Thunderbolt transport node and hot-plug
+    ///     detect asserted, or nil when the machine exposes no transport nodes at all.
+    ///     nil is "the signal is not available here", not "nothing is plugged in":
+    ///     capping on it would black out a desk this rule has never seen. A machine that
+    ///     does expose the nodes and reports none asserted is the #112 state, and 0 is
+    ///     the right answer there.
+    ///
+    /// The cap is an upper bound. It never adds a display, so a dock with a spare socket
+    /// reads the same as one without.
+    static func activeCount(builtin: Int, external: Int, portCap: Int?) -> Int {
+        guard let portCap else { return builtin + external }
+        return builtin + min(external, portCap)
+    }
+}
--- a/Crisp/Models/PhantomPortCap.swift
+++ b/Crisp/Models/PhantomPortCap.swift
@@ -11,9 +11,13 @@ import Foundation
 /// carrying one.
 enum PhantomPortCap {
     /// - Parameters:
-    ///   - builtin: active displays that are the built-in panel. Never capped: the panel
-    ///     is not on a port, so what the ports say has nothing to do with it.
-    ///   - external: active displays that arrived through a port.
+    ///   - offPort: active displays that no port carries: the built-in panel, and a
+    ///     display WindowServer draws for a virtual device (a DisplayLink dock's USB
+    ///     framebuffer, marked kCGDisplayIsVirtualDevice). Never capped: what the ports
+    ///     say has nothing to do with them. Measured on a DisplayLink dock with the lid
+    ///     closed: the machine exposes a transport node for its own HDMI port only, so
+    ///     capping that display read 0 with the screen lit.
+    ///   - onPort: active displays that arrived through a port.
     ///   - portCap: ports with a DisplayPort or Thunderbolt transport node and hot-plug
     ///     detect asserted, or nil when the machine exposes no transport nodes at all.
     ///     nil is "the signal is not available here", not "nothing is plugged in":
@@ -23,8 +27,8 @@ enum PhantomPortCap {
     ///
     /// The cap is an upper bound. It never adds a display, so a dock with a spare socket
     /// reads the same as one without.
-    static func activeCount(builtin: Int, external: Int, portCap: Int?) -> Int {
-        guard let portCap else { return builtin + external }
-        return builtin + min(external, portCap)
+    static func activeCount(offPort: Int, onPort: Int, portCap: Int?) -> Int {
+        guard let portCap else { return offPort + onPort }
+        return offPort + min(onPort, portCap)
     }
 }
--- a/Crisp/Models/PhantomPortCap.swift
+++ b/Crisp/Models/PhantomPortCap.swift
@@ -11,12 +11,12 @@ import Foundation
 /// carrying one.
 enum PhantomPortCap {
     /// - Parameters:
-    ///   - offPort: active displays that no port carries: the built-in panel, and a
-    ///     display WindowServer draws for a virtual device (a DisplayLink dock's USB
-    ///     framebuffer, marked kCGDisplayIsVirtualDevice). Never capped: what the ports
-    ///     say has nothing to do with them. Measured on a DisplayLink dock with the lid
-    ///     closed: the machine exposes a transport node for its own HDMI port only, so
-    ///     capping that display read 0 with the screen lit.
+    ///   - offPort: active displays that no port carries: the built-in panel, and an
+    ///     external that still carries its product name while no port carries it (a
+    ///     DisplayLink dock's USB framebuffer). Never capped: what the ports say has
+    ///     nothing to do with them. Measured on a DisplayLink dock with the lid closed:
+    ///     the machine exposes a transport node for its own HDMI port only, so capping
+    ///     that display read 0 with the screen lit.
     ///   - onPort: active displays that arrived through a port.
     ///   - portCap: ports with a DisplayPort or Thunderbolt transport node and hot-plug
     ///     detect asserted, or nil when the machine exposes no transport nodes at all.
--- a/Crisp/Services/DisplayLuminanceService.swift
+++ b/Crisp/Services/DisplayLuminanceService.swift
@@ -3,7 +3,9 @@ import CoreGraphics
 import IOKit
 import os
 
-private let _CoreDisplayCreateInfoDictionary: (@convention(c) (CGDirectDisplayID) -> Unmanaged<CFDictionary>?)? = {
+/// CoreDisplay's per-display info dictionary. Shared with PhysicalDisplayToggleService,
+/// which reads kCGDisplayIsVirtualDevice from it.
+let _CoreDisplayCreateInfoDictionary: (@convention(c) (CGDirectDisplayID) -> Unmanaged<CFDictionary>?)? = {
     guard let handle = dlopen("/System/Library/Frameworks/CoreDisplay.framework/CoreDisplay", RTLD_LAZY),
           let symbol = dlsym(handle, "CoreDisplay_DisplayCreateInfoDictionary") else { return nil }
     return unsafeBitCast(
--- a/Crisp/Services/PhysicalDisplayToggleService.swift
+++ b/Crisp/Services/PhysicalDisplayToggleService.swift
@@ -168,7 +168,7 @@ final class PhysicalDisplayToggleService: ObservableObject {
         var ids = [CGDirectDisplayID](repeating: 0, count: Int(count))
         guard CGGetActiveDisplayList(count, &ids, &count) == .success else { return 0 }
         let virtual = VirtualDisplayService.shared
-        return ids.prefix(Int(count)).filter { id in
+        let viewable = ids.prefix(Int(count)).filter { id in
             guard !virtual.isVirtualDisplay(id) else { return false }
             // Real panels carry 16-bit EDID vendor and product codes. Two kinds of
             // entry enumerate as active without a screen behind them, and both fail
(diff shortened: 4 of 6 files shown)


ncchen99:
Driven now, on the sequence this rule was read off: built-in held disconnected, sleep,
dock pulled while asleep, woken with the built-in keyboard. Branch is this one merged with
#132 on top of main, scratch bundle, both probes running.

**The desk came back on its own in 4.2 s.** The same sequence on plain main was #112's
run 3: 104 s dark, space twice and the power button doing nothing, and it only returned
when the dock went back in.

```
13:40:14      Sleep
13:40:48      DarkWake  due to USB-C_plug ATC0.USBWakeup        <- dock pulled
13:40:59      Wake      due to RTP.keyboard                     <- built-in keyboard
13:41:02.362  no active display, restoring from 2 record(s)
13:41:03.169  enable 1: reported success after 806 ms
```

The two probes disagree at the same instant, which is the whole point of the change.
`phantom-probe.swift`, computing the count as it stands on main:

```
13:40:46.091 * ... ports[dp:HDMI3 hpd=Unknown sink=0 drv=Ready]      <- both USB-C nodes gone
13:40:59.236 * online=1 active=1 COUNTED=1 | 28[9d1/78e6 A] || ... ports[dp:HDMI3 hpd=Unknown sink=0 drv=Ready]
```

`COUNTED=1`, with real 16-bit EDID, and no port carrying it — the old count stands down
here. `predicate-probe.swift` on the same sample:

```
13:40:58.993 * actual=1(b0+e1) cap=0 effective=0 | 28[9d1/78e6] || dp:HDMI3 hpd=Unknown sink=0
```

`cap=0`, so the phantom is capped away and the rescue runs. The empty HDMI3 node is present
throughout with `hpd=Unknown sink=0` and contributes nothing, which is the case node
presence alone would have got wrong.

The built-in exemption is load-bearing, and this run shows why:

```
13:41:02.726 * actual=1(b1+e0) cap=0 effective=1 | 1[610/a05c B] || dp:HDMI3 hpd=Unknown sink=0
```

`cap` is still 0 with the dock out, and the built-in is the only screen. Because it is not
on a port it is not capped, `effective` is 1, and the rescue stops instead of looping.

The dock went back in at 13:41:36 and both externals returned normally.

**One correction on my own probe**, so nobody is misled by it: `predicate-probe.swift`
labels `effective == 0 while actual > 0` as DANGER, and it printed 17 of those here — about
3.4 s of them, ending when the built-in came up. In this run that label is wrong: `actual`
is the phantom, so zeroing it is the correct answer. The classifier cannot tell "the rule
zeroed a live desk" from "the rule zeroed a phantom desk". It is a useful oracle only on a
run where the displays are known-live, which is what the 1893 samples on #112 were. Worth
saying because I quoted "0 DANGER" as a safety result there.

Files as before — probe output from both, the `com.crisp.app` log and pmset for the window.
Say which and I will paste or attach.

ncchen99:
Covered a connection type none of the runs had touched: moved the second external off the
dock and onto the **Mac's own HDMI port**. Two results, one of which changed the code.

**The built-in HDMI port needs nothing special.** Empty it reads `hpd=Unknown sink=0` in
every sample, which is what made it the example of why node presence alone is not enough.
With a display on it, it reads `hpd=High sink=1` like any USB-C port:

```
dp:HDMI3 hpd=High sink=1; dp:USB-C1 hpd=High sink=1; dp:USB-C2 hpd=Low sink=0
                                                     ^ the dock port the display left
```

Worth noting the emptied dock port settles at `hpd=Low sink=0` rather than `Unknown` — two
different readings for "nothing here", both correctly excluded.

**hpd alone was too strict, so the commit now takes hpd High *or* a counted sink.** With
hpd only, two of the CAPPED samples in that run were a live display's own port:

```
14:24:03.562  CAPPED  cap=1  dp:HDMI3 hpd=Low  sink=1   <- right after an enable
14:24:50.157         ok      dp:HDMI3 hpd=High sink=0   <- during a wake
```

Each signal drops out on its own for a few hundred milliseconds around a transition while
the other holds. Neither is present on an empty port, and in the phantom state the node is
gone entirely, so the pair still reads zero where it has to — that is the property the OR
must not break, and it does not.

**One transient the OR does not fix, and I would rather you knew about it than not.** There
is a window of about 400 ms at display-sleep entry where a port reports *both* signals gone
while CG still lists the display:

```
14:26:55.232  CAPPED  actual=2 cap=1 effective=1 | 48[9d1/78e6] 49[472/e2b]
              || dp:HDMI3 hpd=Low sink=0; dp:USB-C1 hpd=High sink=1
```

Harmless here because the other port holds and the guard only acts at zero. On a
**single-external desk** the same 400 ms would put the cap at 0 with a live display, which
is a wrong rescue. I have not caught that happening — 925 samples on a BenQ-only desk in the
earlier run were clean — but I cannot say it is impossible.

What covers it is #132: the settle check plus ten re-polls spans about a second, so a 400 ms
blank cannot be the sample that decides. The two land better together than either does
alone, which is worth knowing if you were thinking of taking one first.

**Still not covered by any run**: Thunderbolt (`IOPortTransportStateCIO`), which has never
appeared on this desk — the dock's two displays are both `IOPortTransportStateDisplayPort`
and `tbSwitches` never moved. And DisplayLink-class docks, where the display is a USB
framebuffer with no DP transport node at all; on a Mac with a built-in HDMI port the node
count is never zero, so `liveDisplayPortCount()` would return 0 rather than nil and cap a
real screen away. I have no such hardware to check it on. If you would rather the rule
stood down in that shape I can add it, but I could not find a signal that separates it from
the #112 phantom state — both read "nodes present, none asserted, CG says a display" — so I
have left it as measured rather than guessing.

`make check` green.

didriksg (changes requested):
The rule is right for the state it was read off, and the driven run with the dock pulled is exactly the evidence I wanted. But the DisplayLink shape you flagged as uncovered is my own desk, so I ran your arithmetic on it this morning: one ultrawide on a DisplayLink dock, lid closed. The machine exposes one transport node (the built-in HDMI port, hpd Low, sink 0) and nothing for the dock, so the cap is 0 and the count reads 0 with the display lit, four samples running. With that count the Disconnect row disappears from every card once the lid is open, a remembered built-in disconnect gets refused and dropped at launch, and the rescue fires on every callback as soon as a record exists. So as written this cannot merge.

There is a signal that separates it. CoreDisplay_DisplayCreateInfoDictionary for the DisplayLink display carries kCGDisplayIsVirtualDevice = 1 and kDisplayTransportType = 0; it is a USB framebuffer, not a display on a port, which is the same reason the built-in is exempt. If the cap only applies to externals that are not virtual devices, this desk reads 1 and your phantom should still read 0, since it is a DCP display re-enumerated with its EDID. Could you check that key on the phantoms in your capture (the probe can print it alongside hpd), and add the exemption if it holds? I can confirm the DCP side here once a direct display is back on this desk.

PhantomPortCap.activeCount can take the exempt displays as part of builtin or grow a third parameter, either is fine, but the tests should carry the DisplayLink case so it stays covered.

One more thing, and it is about the path rather than this PR. Since #91 the rescue has grown a layer per desk shape: the poll, the shape filter, the settle, the re-polls, now the ports. Each was measured and each was right, but the contract was never written down, so I am writing it here and putting it in the code comment with this change: the rescue exists to bring back a screen that Crisp itself turned off when nothing viewable is left; it reads CoreGraphics and, for the undock case, the port nodes; a display that is not on a port, the built-in or a virtual device, is never doubted; and it accepts about three seconds of dark. This is the last layer. A desk shape it does not cover waits for a report from a release build, not for a new predicate. Same reason #92 waits until 1.6.0 has been out for a couple of weeks; that is not a judgement on the idea.

didriksg:
I have done the two halves I could do from here and pushed them onto your branch, since it allows maintainer edits: a merge of main (the info dictionary loader lives in a file #130 added), and one commit on top. A display that CoreDisplay marks kCGDisplayIsVirtualDevice now sits on the built-in's side of the cap, PhantomPortCap's parameters are named for what they are (off-port and on-port), the tests carry the DisplayLink desk, and the rescue's doc comment states the contract from my review. Verified on the desk that failed this morning: the dock's display reads virtual device, the count reads 1 with the lid closed, four samples.

One thing I changed on the way, and I would rather you saw it than found it: the info dictionary read costs 2.5 to 8 ms per display here, and physicalActiveDisplayCount runs from PhysicalDisplayToggleView's body for every card, so the dictionaries are only read once the ports report fewer than the externals lit. An ordinary desk never pays for it; the phantom state and a virtual-device desk do.

What is left is yours, and it is the half that makes the exemption safe: the phantoms in your capture must not carry that flag. If phantom-probe prints kCGDisplayIsVirtualDevice next to hpd on the undock run, one line per display in the phantom window settles it. Merge once that reads 0 and the build is green.

ncchen99:
Checked it, and the exemption does not hold here. **The phantoms carry
`kCGDisplayIsVirtualDevice = 1`**, so as pushed this branch stands the rescue down and the
desk stays dark — I drove the undock on your commit and it did not come back until the dock
went in.

Your build, `d59b4bc`, scratch bundle, both externals lit, built-in held disconnected:

```
17:55:17  Display is turned off
17:55:17.699 * COUNTED=2 | 53[9d1/78e6 A virt=v] 54[472/e2b A virt=v]
             || ports[USB-C1 hpd=High sink=1; USB-C2 hpd=High sink=1]
17:55:47  Sleep
17:56:05  DarkWake  USB-C_plug                          <- dock pulled
17:56:15  Wake      RTP.keyboard
17:56:15.885 * COUNTED=2 | 53[9d1/78e6 A virt=v] 54[472/e2b A virt=v]
             || ports[dp:HDMI3 hpd=Low sink=0]           <- nothing carrying them
17:56:17.678   restore asked with 1 record(s): 2 active display(s), standing down
             ... dark until the replug at 17:57:09 ...
```

Note the line at **17:55:17**, before the machine slept and with both cables in: two real,
lit monitors on live ports already reading `virt=v`. So this is not a phantom-only
property.

**It is not about virtual devices at all — it is about re-enumeration**, and it reproduces
in forty seconds with no cable touched. `pmset displaysleepnow`, watching the info
dictionary:

```
18:00:16  2[9d1/78e6  virt=0 transport=1  "BenQ GW2780"]  3[472/e2b virt=0 transport=1 "E241Y G0"]
18:00:20  55[9d1/78e6 virt=1 transport=0  name ""]        56[472/e2b virt=1 transport=0 name ""]
18:00:55  2[9d1/78e6  virt=0 transport=1  "BenQ GW2780"]  3[472/e2b virt=0 transport=1 "E241Y G0"]
```

A real DisplayPort monitor, asleep, on a live port, reports `kCGDisplayIsVirtualDevice = 1`
**and** `kDisplayTransportType = 0` **and** an empty product name — the exact signature you
described for the DisplayLink display. Both keys you proposed are the same in both states.
The `unkn/virt` placeholder reads the same too.

Which makes sense of it: the dictionary reports what CoreDisplay knows *now*, and while an
entry is re-enumerated it knows nothing, so it falls back to the generic shape. The phantom
is that state, made permanent by the cable being gone.

**One candidate left, and it is yours to check.** The one field that differed is
`DisplayProductName`: populated normally, empty in the re-enumerated state. If your
DisplayLink display keeps a real product name while it is lit, that separates the two where
the two flags cannot. I have not verified it in the true phantom window — my probe printed
the flags but not the name on the undock run — so before anything is built on it I would
want both halves: the name on your DisplayLink desk, and the name on my phantoms on the
next pull. Happy to do my half whenever.

If it does not hold either, then by your own contract note this desk shape is one the
predicate cannot cover, and the honest options are to take the cap without the exemption
and let a DisplayLink report come from a release build, or to leave #112 open. I would not
guess a third key.

`run14` probe output, the `com.crisp.app` log for the window, and the info-dictionary watcher
are here if you want them.

didriksg:
Pushed one more commit onto your branch, and I would rather close this here than trade another round of captures. Good catch on the re-enumeration state; my half reads the way you hoped: with the lid closed the lit DisplayLink display carries the same two flags as your phantoms and a full DisplayProductName ("Dell U4919DW" in every locale), so the name is populated exactly where the flags are not enough.

Two things in 6fc5777. The exemption is the product name now instead of kCGDisplayIsVirtualDevice, on your measurement that the re-enumerated entry empties it. And the cap is read by the rescue only: the Disconnect row and the launch re-apply keep the plain count, so if the ports are ever wrong on a desk neither of us has seen, the worst case is the rescue firing, which brings back a screen Crisp turned off, and never a hidden row or a refused disconnect.

I did not wait for the name on your phantoms, and here is why that is safe: if a phantom turns out to carry a name it is counted and the rescue stands down, which is exactly what main does today, so being wrong costs you nothing you do not already have and costs everyone else nothing. The other residual is written in the doc comment too: a DisplayLink-only desk in display sleep reads zero for the length of the sleep, and with a record on the books the rescue would fire there. Your next pull is a field report, not a gate. Merging when the build is green; #112 closes with it.

didriksg (approved):
Approving over my earlier changes-requested: the port cap is scoped to the rescue and the exemption is the product name, per the comment above.

ncchen99:
The pull happened, so: field report, as you said, not a gate. **The phantom carries no product name**, so the exemption lands where you put it.

Build is main at `29a0223` with my #92 branch on top, in a scratch bundle — saying so because it is not plain main, though nothing it adds sits on this path and the restore below is upstream's. BenQ on the dock's DisplayPort as the only lit screen, built-in held disconnected. Sleep, pull the dock, wake, leave it alone:

```
20:08:55      display off, id 2 -> 72, DisplayProductName goes empty
20:09:00      Sleep
20:09:30      DarkWake  USB-C_plug                     <- dock out
20:09:29.037  cap=0 effective=0, CG still lists 72[9d1/78e6]
20:09:51      Wake  RTP.keyboard
20:09:56.985  no active display, restoring from 2 record(s)
20:09:57.057  enable 1: reported success after 71 ms
```

Six seconds from the wake. The run that opened #112 was 104 s and only came back on the replug; the virtual-device build never came back at all.

The name half is confirmed by Crisp rather than by my probe, which is the stronger way round: the merged code exempts a named display from the cap, so `no active display, restoring` is only reachable if `hasProductName(72)` returned false. The count read 0 on an entry CoreGraphics was still reporting as a real BenQ. My watcher agrees — the name emptied at display sleep and never came back until the built-in replaced the entry — but it prints only on change and has no heartbeat, so I would lean on the in-app line rather than on it.

Both records survived and the desk ended where it started.

That closes the first residual in the doc comment on this desk. The second, a DisplayLink-only desk in display sleep, I still have no hardware for.

PR #165, opened 2026-09-10
Add optional system Night Shift temperature slider

Anywhere-Music-Player:
Crisp can toggle Night Shift, but adjusting its system warmth still requires opening macOS System Settings. This adds an optional **Night Shift Temperature** slider below the system effects buttons and above Presets.

Years ago I added this control to my [NativeDisplayBrightness](https://github.com/Anywhere-Music-Player/NativeDisplayBrightness) app. After moving to Crisp for multiple displays, I wanted that same convenient access to the system Night Shift setting, which motivated this PR.

- Enable it with **Settings → Show Night Shift Temperature** (off by default).
- The slider uses **Less Warm / More Warm** endpoints and is disabled while Night Shift is off. The existing Night Shift button remains the on/off control.
- It reads and writes system strength through the existing CoreBrightness client, with runtime selector checks. macOS continues to own the temperature and schedule; Crisp persists only visibility.
- Writes run off the main thread, are serialized, and coalesce to the latest value. Stale reads cannot overwrite an active drag. External changes are picked up by the existing status, panel-open, and visible-panel refresh paths; strength-only changes do not reliably emit a status callback on the tested Mac.
- The control uses a separate panel canvas block and includes English and Simplified Chinese strings.
- Align the generated Xcode project with the Swift 5 language mode already used by the dev and CI builds, and remove the language overrides from `make test` and `make loc-check` so those checks exercise the project default. The unchanged base commit also fails a default Swift 6 build in ArrangementView, BrightnessSliderView, and UpdateService.

Validation:

- `make check`: strict SwiftLint, application build and 142 passing unit tests, x86_64 typecheck, and localization key validation, using the project language mode without a command-line override.
- Build and all seven Night Shift controller tests passed in the open Xcode 27 RC workspace, using local ad-hoc signing and the project language mode. The PR does not change project signing defaults.
- Seven new controller tests cover external changes, failed reads/writes, invalid values, stale reads, dragging, and write coalescing.
- On macOS 26.6.2, verified the runtime method signatures and actual CoreBrightness write/readback, then restored and verified the original system strength. Also verified synchronization through the existing refresh path after a separate client changed strength.
- Rendered the actual SwiftUI slider component offscreen at the default 320-point panel width.

Full-panel mouse interaction and sleep/wake behavior have not been manually verified. Smooth brightness stepping is outside this PR.

### DIFF
--- a/Crisp/App/AppDelegate.swift
+++ b/Crisp/App/AppDelegate.swift
@@ -719,6 +719,11 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
             || CoreBrightnessService.shared.trueToneAvailable {
             blocks.append(block("effects") { ScreenEffectsView() })
         }
+        if let temperature = CoreBrightnessService.shared.nightShiftTemperature {
+            blocks.append(block("nightShiftTemperature", isOpen: { settings.showNightShiftTemperature }) {
+                NightShiftTemperatureView(temperature: temperature)
+            })
+        }
         blocks.append(block("presets") {
             VStack(alignment: .leading, spacing: 0) {
                 SectionDivider()
@@ -800,6 +805,10 @@ class AppDelegate: NSObject, NSApplicationDelegate, NSWindowDelegate {
             Task { @MainActor in self?.canvas.refreshAppearance() }
         }
         SettingsService.shared.$showCombinedBrightness
+            .dropFirst()
+            .sink { [weak self] _ in self?.canvas.requestApply() }
+            .store(in: &canvasCancellables)
+        SettingsService.shared.$showNightShiftTemperature
             .dropFirst()
             .sink { [weak self] _ in self?.canvas.requestApply() }
             .store(in: &canvasCancellables)
--- a/Crisp/Models/NightShiftTemperatureController.swift
+++ b/Crisp/Models/NightShiftTemperatureController.swift
@@ -1,0 +1,54 @@
+import Foundation
+import Combine
+
+/// The system owns the strength and schedule. Keep only a live UI snapshot here.
+/// Reads cannot pull the thumb backward during a drag; writes are serialized and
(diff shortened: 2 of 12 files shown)


didriksg:
Thanks for this, and it's nice to hear you moved over to Crisp for your displays. The controller is careful work: a stale read can't pull the thumb back during a drag, writes coalesce to the latest value, and the selectors are checked at runtime. The tests cover exactly the cases that would go wrong.

Still, I'm going to pass on this one. Crisp tries to stay lightweight, as an expansion of what Control Center offers for displays. Control Center has the Night Shift button but no warmth slider, since macOS keeps that in System Settings and most people set it once. Crisp also already has a warmth control, the Color Temp row in each display's image adjustment, and a second warmth slider that works differently (every display, and only while Night Shift is on) would be hard to tell apart from it.

The Swift 5 language mode change is right, though. Every build Crisp runs already uses Swift 5 mode, and only a build from inside Xcode picks up 6.0 and fails. I'll take your commit from #166 into a small PR of its own, with you as the author.

If more people ask for a warmth slider, I'll come back to this PR.

Anywhere-Music-Player:
Thanks for the thoughtful review and the feedback on the controller and tests.

My use case is quick access to the system Night Shift warmth without opening System Settings each time. The per-display Color Temp adjustment serves a different purpose for me, so it doesn't replace this control in my workflow.

I understand that you want to keep Crisp lightweight. I'll keep this feature in my fork for now. I'm planning a dedicated, modern settings window for my fork, where I can organize these options more clearly and keep the menu bar compact. That might also help distinguish system Night Shift settings from per-display image adjustments.

Thanks for reviewing this. The PR is here if you decide to revisit it.

didriksg:
Sounds like a good idea! Feel free to come up with suggestions and things that work from your fork in the future! I'm always open to suggestions and features that fit into Crisp's native ambition! :)

PR #171, opened 2026-09-16
Let a display off the software brightness latch when DDC answers again (#167)

### LINKED ISSUE #167: Two identical external monitors (U32N10) look different in brightness at the same slider value
### What happened

I have two identical external monitors (both **U32N10**). In Crisp, both brightness sliders are set to the same value (both at max, and the "Combined brightness" slider is at max too), but the two screens are **visibly different in brightness**. I expected two identical monitors at the same slider value to look the same.

Crisp shows them as `U32N10(1)` and `U32N10(2)` with identical rows. Nothing in the menu shows that the two are driven differently.

### What the log shows

Pairing is fine: both displays were matched to their own DDC channel by identity.

```
09:50:22.594 [ddc] pairing: display 2 (vendor 0x05E3 product 0x3210) -> channel 1 of 2 by identity
09:50:22.594 [ddc] pairing: display 5 (vendor 0x05E3 product 0x3210) -> channel 0 of 2 by identity
```

Display 5's DDC worked, then its I2C channel started failing and never recovered:

```
13:13:21.902 [ddc] read 0x10 display 5: ok 8/100          <- last good hardware value
13:13:39.226 [ddc] read 0x10 display 5: reply read I2C error 0xE0114101
13:14:09.807 [ddc] display 5: 6 consecutive read failures, reads quarantined for 600 s
... every re-probe (13:24, 13:34, 13:52) fails with 0xE0114101, ~6 s per attempt ...
13:42:31.824 [ddc] write 0x10=7 display 5: failed all 3 attempts
13:42:50.269 [ddc] write 0x10=56 display 5: failed all 3 attempts
13:53:17.258 [ddc] write 0x10=76 display 5: failed all 3 attempts
13:53:17.258 [brightness] display 5: 3 consecutive DDC brightness writes failed, brightness now software gamma until reconnect
```

Display 2 keeps reading fine the whole time and is at `100/100` over DDC.

`crispctl display list` confirms the split:

```
U32N10(1)  id 2  brightnessBackend "ddc"       brightness 100
U32N10(2)  id 5  brightnessBackend "software"  brightness 100
```

So the mismatch is:

- `U32N10(1)`: hardware backlight 100 (DDC).
- `U32N10(2)`: hardware backlight stuck at its last value (last read 8/100), gamma at 100 (identity). Its slider says 100, but the panel is really at its old backlight level.

### Why Crisp stays stuck

- `ddcWriteFailed` latches `ddcAvailable[id] = false` after 3 failed writes (`BrightnessService.swift`, around line 789).
- Once latched, the refresh path skips DDC reads completely (`knownUnavailable`, around line 349), and writes go straight to gamma. The only thing that clears it is `invalidateDDCTopology` (a display reconfiguration, e.g. a replug). So if the I2C channel comes back, Crisp never notices.
- In software mode the slider shows the gamma level, not the real backlight, and the menu gives no sign that this display is not over DDC.

The I2C failure itself (0xE0114101) is outside Crisp. Crisp's part is the permanent latch and the slider showing 100 while the backlight is not.

### Steps to reproduce

1. Connect two identical external monitors over HDMI (here 2× U32N10).
2. Use Crisp's brightness on both until one display's DDC channel starts failing (here it began on its own after ~3 h; unplugging that monitor's DDC/HDMI mid-session should also force it).
3. Keep adjusting brightness until Crisp logs `3 consecutive DDC brightness writes failed`.
4. Set both sliders to the same value: the screens no longer match, and the failed one stays in software mode even after DDC works again.

### Crisp version

1.6.0

### macOS version

macOS 26.4 (25E246)

### Mac model and chip

MacBook Pro (Mac17,8), Apple M5 Pro

### Monitor(s) and how they are connected

2× U32N10 (3840×2160), both over **HDMI**. One in landscape, one rotated 90° (portrait). Built-in Liquid Retina XDR also on.

### Anything else

Dark mode on, Night Shift off, True Tone on.

I'll take this one: I'm looking into the cause and plan to open a PR with a fix.

DEOWL-kan:
Closes #167

## What this changes

Three failed DDC writes latch an external display to software gamma. Until now only a reconnect cleared that latch, so the OSD-menu recovery in `docs/ddc-notes.md` went unnoticed: the slider kept showing the gamma level while the backlight sat at its last hardware value, and two identical monitors at the same slider value looked different.

- A latched display gets **one DDC read per panel open**, and only if it answered a read since it connected, is not HDR-dimmed, and had no brightness change land during that read.
- The read alone does not unlatch it. The latch is about writes, so the read earns a **verification write carrying the slider's current value**. The existing write-success path unlatches, clears the streak and settles the gamma dim. A write that fails, or that a drag supersedes, leaves the latch and the picture untouched; the next panel open probes again.
- The 1 s poll never reads a latched display: a reconfiguration flushes every read quarantine, and polling a dead channel then ties up its queue for tens of seconds.
- `readAsync` gains `probeQuarantined`, which lets one raw read through a **live** quarantine (streak one short of the threshold, so a failure re-quarantines at once). It is lifted on the display's own queue, right before the read, so a read queued ahead — the panel-open volume probe — cannot spend the attempt. An expired quarantine is left to `readSynchronous` so its fresh window survives.

## Why the write, and not the read, decides

Unlatching on a read alone was the first shape of this fix, and it was wrong in two ways. It jumped the slider to the stale backlight, so a display latched at backlight 80 with the slider dialled to 20 leapt back to 80 on a panel open the user asked nothing of. And on a display whose reads work but whose writes are NACKed (picture-mode or OSD brightness locks do this), that jump repeated on every open. Writing what the slider already shows keeps the level the user picked, makes two identical monitors actually match, and proves the channel takes writes before anything visible changes.

## Impact

`BrightnessService` (read plan, read handling, write-success unlatch guard), `DDCService` (`readAsync` probe parameter), `AppDelegate` (panel-open refresh passes `probe: true`), plus `docs/ddc-notes.md`. Behaviour for a display that is not latched is unchanged: the new guard in `ddcWriteSucceeded` short-circuits on both other states, and `ddcReadPlan` returns the same answer it always did.

Cost on a display that stays dead: one raw read per panel open while quarantined (~6 s on that display's own serial queue), plus one failed verification write once it answers reads again. A `hold()` around a display enable/disable waits behind that.

## Verification

Two identical U32N10 over HDMI on an M5 Pro, macOS 26.4, with local fault injection in `arm64Read`/`arm64Write` (never committed):

| Case | Result |
| --- | --- |
| Three failed writes latch the display | PASS |
| One probe read per panel open, exactly | PASS (5 opens, 5 raw reads) |
| Volume read queued ahead cannot spend the probe | PASS (it fails fast on the quarantine, the probe still reads) |
| Recovery keeps the slider value, hardware takes it | PASS (backlight 80 + slider 20 → slider stays 20, hardware reads 20, gamma 0.2 → 1) |
| Backlight below the 15% blend threshold | PASS (backlight 8 → gamma 0.533, no visible step) |
| Drag during the probe read | PASS (no unlatch; next open recovers) |
| Drag during the verification write | PASS (no unlatch; state stays consistent; next open recovers) |
| Read-ok / write-dead display, three opens | PASS (stays latched, gamma unchanged, no flapping) |
| Display that never answered a read | PASS (never probed) |
| Auto-brightness on | PASS (recovery only on panel open) |
| Expired quarantine | PASS (fresh full window, expiry logged) |
| Sleep/wake, physical replug, HDR, Extra Brightness, colour temperature | SKIP — not constructible here or needs a person; reasons recorded |

`make check` green (SwiftLint strict, unit tests, x86_64 typecheck, localization keys).

No unit test was added: the new decisions live behind locks in a singleton, so there is no seam the existing test target can reach. They are covered by the device runs above and by a log line on every outcome, including the two silent ones (verification write superseded, verification write failed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Ufn6ZWjgfE7esNtTjjKM8u


didriksg:
The reasoning about the latch is right. A good read proves nothing about writes, and writing the value the slider already shows is a better recovery than jumping to the stale backlight. Your fault-injection runs show the code does what it says.

What they can't show is whether a deaf channel ever comes back without a reconnect, and this PR only helps if it does. That's the question from my comment on #167, and I'm sorry to hold this up after so much work. Your log can't answer it either: after the latch at 13:53 nothing read display 5 again until you replugged at 14:28, so there's no way to tell whether the channel came back before that.

Next time U32N10(2) goes deaf, could you try these before you touch the cable?

1. Read brightness with m1ddc. If it doesn't answer, open the monitor's OSD menu for a few seconds, close it, and read again.
2. If it's still deaf, use Disconnect and then Reconnect on U32N10(2) in Crisp's menu, and look for a "DDC brightness read ok" line in the log.

Each answer points to a different fix. If m1ddc answers in step 1, Crisp has to notice a channel that comes back on its own, which is what this PR does, and I'll review the code properly then. If only Disconnect and Reconnect work, the latch already clears on a reconnect today, so the fix is for Crisp to point you to Reconnect. If only a replug works, the menu should show that the display lost hardware brightness, since Crisp can't move the backlight without DDC.

I'll leave the PR open until we know.

DEOWL-kan:
You're right, and I should have waited for that measurement. You asked for it on #167 before any code, and I missed the comment — I posted the replug log and went straight into the fix without checking back. That's on me, not on you for holding this up.

On the evidence itself, your reading of my log is correct: nothing read display 5 between the latch at 13:53 and the replug at 14:28, so it cannot say whether the channel came back on its own. I also can't recover that window now — the unified log store on this Mac only goes back to 15 Sep, so the original capture is gone beyond the excerpts already quoted in #167. It hasn't gone deaf again since.

So: next time U32N10(2) goes deaf, before the cable, I'll run your three steps and report the raw output — brightness read, then read again after opening and closing the monitor's OSD menu, then Disconnect/Reconnect in Crisp's menu with the log line either present or absent. I'll use `scripts/ddc-probe.swift` from this repo for the reads rather than m1ddc, since it reads VCP 0x10 per channel and dumps the raw reply bytes; say the word if you'd rather have m1ddc's output for comparability.

Happy to leave this PR sitting until then, and to close it without ceremony if the answer turns out to be "only a replug works" — in that case the useful change is the one you describe, telling the user the display has lost hardware brightness, which is a different patch. Thanks for reading it as carefully as you did.

didriksg:
ddc-probe.swift is a better pick than m1ddc: it pairs channels the way Crisp does and prints the raw reply, so no reply and a garbage reply look different. Two things keep the result clean. Run it on 1.6.0 rather than this PR's build, since the PR reads and writes on every panel open and would mix Crisp's own traffic into the test. And keep Crisp's panel closed while the script runs. The two monitors are the same model, so paste Crisp's pairing line from the same session too (`display 5 ... -> channel 0 of 2`), so I can tell which channel is U32N10(2).

PR #175, opened 2026-09-19
Volume keys and the slider's ± buttons only move on the 16-stop grid

Yangmoooo:
## What & why

#172 gave the brightness keys a quarter step with Option+Shift held (`BrightnessKeySteps.fineStep`), but the volume path never picked it up:

- `routeVolumePress` calls `BrightnessKeySteps.next(from:up:)` without `fine:`, so a volume key press always moves a full 1/16 stop.
- `VolumeSliderView` hardcodes the same full step (`volumeStep = 100/16`), even though the comment right above it says it matches the volume keys.

That hurts on a monitor that only takes DDC volume writes. My LG 27UP850 (4K over USB-C/DP audio — the "never answers a 0x62 read but accepts writes" case the VolumeService comment already names for the 27UP850N) sits in `crisp.volumeForcedDisplays`, so one key press steps the DDC value by 100/16 ≈ 6.25: 0 → 6 → 13. The monitor is loud enough that the usable range is roughly 0–15, so every press jumps several audible steps, and the slider is the only control that can land on a value.

Two commits:

1. Pass the same `fine` flag to the volume keys. The model already supports it, so that is the whole change:

       let fine = event.flags.contains([.maskAlternate, .maskShift])
       ...
       case Self.nxKeytypeSoundUp:
           service.setVolume(BrightnessKeySteps.next(from: target.volume, up: true, fine: fine), for: target)
       default:
           service.setVolume(BrightnessKeySteps.next(from: target.volume, up: false, fine: fine), for: target)

2. Let the slider's ± buttons move the fine step too (`BrightnessKeySteps.fineStep`, comment updated). Tapping them repeatedly is how you nudge a write-only display by a point or two. If you would rather keep the modifier convention keyboard-only, Option-click on the buttons would do instead — happy to drop this commit.

Mute is untouched, and no new user-facing string is added.

## How tested

- `swiftlint lint --strict` — clean.
- Universal Debug build (arm64 + x86_64) of `Crisp`, `crispctl` and `CrispTests` via `xcodebuild` — clean.
- `xcodebuild test` — 139 tests, 0 failures.
- `xcodebuild -exportLocalizations` + `scripts/check-localization-keys.py` — clean, 221 keys, no new ones.
- Environment note, unrelated to this change: this machine runs Xcode 27, where the CLT-only paths don't run — `make compile`/`make typecheck-x86` because raw `swiftc` can't find the SwiftUI macro plugin, and `make test` because the generated project sets no `PRODUCT_NAME` for the app and test targets (empty module name). I patched the generated project locally (it is gitignored) and ran the suite with `ARCHS=arm64` and warnings-as-errors off, since `DisplayManager.swift:120` now trips the new `ImplicitStrongCapture` diagnostic. CI stays the gate; happy to file that separately if it is useful.
- Not yet measured on hardware: with the patch a press with Option+Shift moves a quarter stop (1.5625%), which the DDC writer rounds to 1–2 units of the display's 0–100 range instead of 6–7.

## Checklist
- [ ] Added/updated `Crisp/Resources/Localizable.xcstrings` for any new user-facing strings — n/a, no new strings
- [ ] Screenshot or short clip for any UI change — n/a, the buttons' step size changes, nothing visual


didriksg:
Thanks, this is a good catch. #172 should have covered the volume keys, and your first commit fixes that. It reads the flags the same way the brightness path does, and the bezel on 14 and 15 already draws the quarter steps for volume.

I kept the ± buttons at the full step. A tap should move the same distance as a plain key press, and on a monitor that answers reads, the quarter step would take 64 taps to go from 0 to 100. With your first commit, your 27UP850 gets its fine control from Option+Shift on the keys, and the slider can already set any value.

So I didn't have to ask you for another round, I pushed to your branch myself. I dropped the second commit, rebased onto main and added your credit line in ACKNOWLEDGMENTS.md.

PR #177, opened 2026-09-21
Hide the panel on the second status item click

Dr8pus:
## What & why

A second click on the status item flashed the panel instead of hiding it. That
press hits one of the panel's auto-dismiss paths before the click interceptor's
mouse-down monitor, so the panel was already hidden when the interceptor
toggled, and it showed it again.

Both auto-dismiss paths (resign-key, and the outside-click monitor) now skip a
press that lands on the status item, so togglePanel is the only thing that opens
or closes the panel.

## How tested

Built and ran the app locally (Debug), then clicked through:

1. Click the status item — the panel opens.
2. Click it again — the panel hides cleanly (this is the case that used to
   flash and stay open).

## Checklist
- [x] Screenshot or short clip for any UI change

https://github.com/user-attachments/assets/5d721ade-cef7-494f-8c91-03a920c567df


didriksg:
I added your credit line to ACKNOWLEDGMENTS.md on your branch. Thanks for tracking this one down.

Dr8pus:
Thanks for adding it. Happy to take another one if anything like this comes up.

